PR #57, opened 2020-03-23
solve exercise1

guysalt:


avrahamshukron:
Your PR does not contain any changes. Did you forget to stage your changes before committing?

avrahamshukron (approved):

exercise1.py
```diff
@@ -0,0 +1,8 @@
+print("enter your age:", end=' ')
+age = int(input())
+if age == 18:
+    print('Congratulations')
```
avrahamshukron, line 4: Please be consistent with your string quoting. Unless you have a good reason not to, stick to either `"` or `'`.
I recommend `"` only.

avrahamshukron (changes requested):

PR #59, opened 2020-03-23
solve exercise2 - fibonacci

guysalt:


avrahamshukron:
Second time is a charm :)

The staging operation in git is probably the most un-intuitive to newcomers.

PR #60, opened 2020-03-24
Matrix Exercise

yonatanevantal:
Pull request to merge matrix exercise to the master.

### DIFF
--- a/matrix/matrix.py
+++ b/matrix/matrix.py
@@ -13,9 +13,7 @@ class Matrix:
     __slots__ = '_matrix'
 
     def __init__(self, *lines):
-        if not self._is_valid_data(*lines):
-            raise ValueError('Data is not valid')
-
+        self._validate(*lines)
         super(Matrix, self).__setattr__('_matrix', tuple(lines))
 
     def __setattr__(self, key, value):
@@ -23,24 +21,22 @@ class Matrix:
         raise AttributeError(f'{self.__class__} instance cannot be modified')
 
     @staticmethod
-    def _is_valid_data(*data: tuple) -> bool:
+    def _validate(*data: tuple):
         """Checks if the data given is valid. Valid data is defined if it
         contains tuples of the same length (Matrix only supports n * n) and
         their value is an integer."""
         # check if each line is a tuple.
         if not all([isinstance(line_data, tuple) for line_data in data]):
-            return False
+            raise ValueError("Matrix lines must be tuple")
 
-        # check if each item in each line is an int.
+        # check if each item in each line is an int or float.
         if not all([any([isinstance(item, int), isinstance(item, float)]) for
                     line_data in data for item in line_data]):
-            return False
+            raise ValueError("Matrix items must be int or float")
 
         # checks if the given data is of a n * n matrix.
         if not all([len(line) == len(data) for line in data]):
-            return False
-
-        return True
+            raise ValueError("Matrix must be even-sided")
 
     @classmethod
     def unity(cls, number_lines: int):
@@ -83,13 +79,8 @@ class Matrix:
             'not equal', return true if one or more items are not equal to
             their parallel item/s in other, else false.
         """
-        if operation == operator.eq:
-            return all(
-                [item for line_data in data for item in line_data])
-
-        elif operation == operator.ne:
-            return any(
-                [item for line_data in data for item in line_data])
+        func = {operator.eq: all, operator.ne: any}.get(operation)
+        return func([item for line_data in data for item in line_data])
 
     def _calculate_matrix_on_matrix(self, other, operation):
         """Does mathematical operation on two matrices.
--- a/matrix/matrix.py
+++ b/matrix/matrix.py
@@ -49,12 +49,8 @@ class Matrix:
     @classmethod
     def unity(cls, dimension: int):
         """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
-        unity_matrix = ()
-
-        for i in range(dimension):
-            unity_matrix += ((0,) * i + (1,) + (0,) * (dimension - i - 1),)
-
-        return cls(unity_matrix)
+        return cls(tuple((0,) * i + (1,) + (0,) * (dimension - i - 1) for i
+                         in range(dimension)))
 
     @classmethod
     def ones(cls, dimension):
--- a/matrix/matrix.py
+++ b/matrix/matrix.py
@@ -37,8 +37,7 @@ class Matrix:
             return False
 
         # checks if the given data is of a n * n matrix.
-        current_length = len(next(iter(data)))
-        if not all([current_length == len(data) for _ in data]):
+        if not all([len(line) == len(data) for line in data]):
             return False
 
         return True
@@ -104,7 +103,7 @@ class Matrix:
         """
         return [tuple([operation(self.tuples[i][j], other.tuples[i][j]) for j
                        in range(len(self.tuples))]) for i in range(len(
-                        self.tuples))]
+            self.tuples))]
 
     def _calculate_scalar_on_matrix(self, other, operation):
         """Does mathematical operation on scalar and matrix.
@@ -117,7 +116,7 @@ class Matrix:
             List of tuples which contains the calculated matrix data.
         """
         return [tuple([operation(self.tuples[i][j], other) for j in range(
-                    len(self.tuples))]) for i in range(len(self.tuples))]
+            len(self.tuples))]) for i in range(len(self.tuples))]
 
     def _operations_on_matrix(self, other, operation):
         """Does all the operation that can be done on a matrix.
@@ -136,22 +135,27 @@ class Matrix:
         """
         if isinstance(other, Matrix):
 
-            if len(self.tuples) != len(other.tuples):
+            if len(self.tuples) != len(other.tuples) and not any(
+                    [operation == operator.eq, operation == operator.ne]):
                 raise ValueError("Matrices length don't match")
 
             matrix_data = self._calculate_matrix_on_matrix(other, operation)
-            
+
             if operation == operator.eq or operation == operator.ne:
                 return self._bool_operation(operation, matrix_data)
-            
+
             return Matrix(*matrix_data)
 
-        elif isinstance(other, int) or isinstance(other, float):
+        elif any([isinstance(other, int) or isinstance(other, float)]) and any([
+            operation == operator.mul, operation == operator.truediv]):
             return Matrix(*self._calculate_scalar_on_matrix(other, operation))
+
+        else:
+            raise ValueError(f"{operation} not supported on {type(other)}")
 
     def __add__(self, other):
         """Adds two matrices."""
-        return self._operations_on_matrix(other, operator.mul)
+        return self._operations_on_matrix(other, operator.add)
 
     def __sub__(self, other):
         """Subtract two matrices."""
--- a/matrix/matrix.py
+++ b/matrix/matrix.py
@@ -1,7 +1,5 @@
 """Matrix data type class."""
 
-import operator
-
 
 class Matrix:
     """An immutable object representing a matrix data-type.
@@ -9,7 +7,7 @@ class Matrix:
     Attributes:
         _matrix (tuple): contains the data about the matrix. Each line of
             the matrix is represented by a tuple.
-    Note:
+    note:
         This implementation only supports squared matrices (n * n).
     """
 
@@ -20,12 +18,12 @@ class Matrix:
     def _validate(self, data: tuple):
         """Check if the data given is valid.
 
-        Note:
+        note:
             Valid data is defined if it contains tuples of the same length
             (Matrix only supports n * n), items can't be iterable.
         """
-        if any(hasattr(item, '__len__') for line in data for item in line):
-            raise TypeError("Matrix item type not supported")
+        # if any(hasattr(item, '__len__') for line in data for item in line):
+        #     raise TypeError("Matrix item type not supported")
 
         if not all([len(line) == len(data) for line in data]):
             raise ValueError("Matrix must be even-sided")
@@ -34,14 +32,13 @@ class Matrix:
     def unity(cls, dimension: int):
         """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
         return cls(
-            tuple([tuple((0,) * i + (1,) + (0,) * (dimension - i - 1)) for i in
-                   range(dimension)]))
+            tuple(tuple((0,) * i + (1,) + (0,) * (dimension - i - 1))
+                  for i in range(dimension)))
 
     @classmethod
     def ones(cls, dimension):
         """Generate a matrix full of 1's."""
-        return cls(tuple(([tuple((1,) * dimension) for _ in
-                           range(dimension)])))
+        return cls(((1,) * dimension,) * dimension)
 
     def __repr__(self):
         return f'{self.__class__.__name__}{self._matrix}'
@@ -55,65 +52,60 @@ class Matrix:
         """Return the matrix instance as a tuple of tuples."""
         return self._matrix
 
-    def _matrix_by_matrix(self, other, operation):
-        """Calculate operation on two matrices.
+    def _add_matrix(self, other):
+        """Add two matrices.
 
         Args:
-            operation (function): the function of the operator.
             other (Matrix): the other matrix to operate on.
 
         Returns:
-            Matrix: the result of the operation.
-
-        Note:
-            Boolean operations supports un-squared matrices but
-            mathematical operations don't.
+            Matrix: the result of the addition.
         """
-        if len(other.tuples) != len(self.tuples) and not any(
-                [operation == operator.eq, operation == operator.ne]):
+        dimension = len(self.tuples)
+
+        if len(other.tuples) != dimension:
             raise ValueError("Matrices length don't match")
 
-        calculated_matrix = []
+        calculated_data = [[0 for i in range(dimension)]
+                           for j in range(dimension)]
 
         for i in range(len(self.tuples)):
-
-            calculated_line = []
             for j in range(len(self.tuples)):
-                calculated_line.append(
-                    operation(self.tuples[i][j], other.tuples[i][j]))
+                calculated_data[i][j] = self.tuples[i][j] + other.tuples[i][j]
 
-            calculated_matrix.append(tuple(calculated_line))
+        calculated_data = [tuple(line) for line in calculated_data]
+        return Matrix(tuple(calculated_data))
 
-        return Matrix(tuple(calculated_matrix))
-
-    def _by_scalar(self, scalar, operation):
-        """Calculate operation on matrix by scalar.
+    def _matrix_mul_matrix(self, other):
+        """Multiply two matrices.
 
         Args:
-            operation (function): the function of the operator.
-            scalar (number): the scalar to operate by.
+            other (Matrix): the other matrix to multiply.
 
         Returns:
-            Matrix: the result of the operation.
-
-        Note:
-            Scalar must be a number. Operations with scalars are only * and /.
+            Matrix: the product.
         """
-        calculated_matrix = []
+        dimension = len(self.tuples)
 
-        for i in range(len(self.tuples)):
+        if len(other.tuples) != dimension:
+            raise ValueError(
+                f"Dimension {len(other.tuples)} don't match {dimension}")
 
-            calculated_line = []
-            for j in range(len(self.tuples)):
-                calculated_line.append(operation(self.tuples[i][j], scalar))
+        calculated_data = [[0 for i in range(dimension)]
+                           for j in range(dimension)]
 
-            calculated_matrix.append(tuple(calculated_line))
+        for i in range(dimension):
+            for k in range(dimension):
+                for j in range(dimension):
+                    calculated_data[i][j] += self.tuples[i][k] * \
+                                             other.tuples[k][j]
 
-        return Matrix(tuple(calculated_matrix))
+        calculated_data = [tuple(line) for line in calculated_data]
+        return Matrix(tuple(calculated_data))
 
     def __add__(self, other):
         """Add two matrices."""
-        return self._matrix_by_matrix(other, operator.add)
+        return self._add_matrix(other)
 
     def __sub__(self, other):
         """Subtract two matrices."""
@@ -122,33 +114,26 @@ class Matrix:
     def __mul__(self, other):
         """Multiply two matrices or scalar by matrix."""
         if isinstance(other, Matrix):
-            return self._matrix_by_matrix(other, operator.mul)
+            return self._matrix_mul_matrix(other)
 
-        return self._by_scalar(other, operator.mul)
+        # if scalar multiplication
+        return Matrix(tuple(
+            [tuple(map(lambda x: x * other, line)) for line in self.tuples]))
 
     def __rmul__(self, other):
-        """Define to handle reverse multiplication (10 * matrix) as regular
-        multiplication."""
+        """Define reverse multiplication as regular multiplication."""
         return self * other
 
     def __truediv__(self, other):
-        """Divide matrix by scalar. Returns floats if
-        needed."""
-        return self * (1 / other)
-
-    def _check_bool_matrix(self, matrix, operation):
-        """Return true if all the items in the matrix are true else false."""
-        func = {operator.eq: all, operator.ne: any}.get(operation)
-        return func(
-            [item for line_data in matrix.tuples for item in line_data])
+        """Divide matrix by scalar. Return floats if needed."""
+        return self * (1.0 / other)
 
     def __eq__(self, other):
-        """Compare two matrices. Returns true if they are equal else false."""
-        return self._check_bool_matrix(
-            self._matrix_by_matrix(other, operator.eq), operator.eq)
+        """Compare two matrices. Return true if they are equal else false."""
+        return self.tuples == other.tuples
 
     def __ne__(self, other):
-        """Compare two matrices. Returns true if they are not equal else
+        """Compare two matrices. Return true if they are not equal else
         false."""
         return not self == other
 
@@ -159,3 +144,8 @@ class Matrix:
     def __hash__(self):
         """Enable to use a matrix as key of a dictionary."""
         return hash(self.tuples)
+
+
+m1 = ((1, 2), (3, 4))
+m2 = Matrix.ones(3)
+print(repr(m2))
--- a/matrix/matrix.py
+++ b/matrix/matrix.py
@@ -1,7 +1,6 @@
 """Matrix data type class."""
 
 import operator
-from collections.abc import Iterable
 
 
 class Matrix:
@@ -14,42 +13,38 @@ class Matrix:
         This implementation only supports squared matrices (n * n).
     """
 
-    def __init__(self, *lines):
-        self._validate(*lines)
-        self._matrix = tuple(lines)
+    def __init__(self, lines):
+        self._validate(lines)
+        self._matrix = lines
 
-    def _validate(self, *data: tuple):
+    def _validate(self, data: tuple):
         """Check if the data given is valid.
 
         Note:
             Valid data is defined if it contains tuples of the same length
             (Matrix only supports n * n), items can't be iterable.
         """
-        if not all(isinstance(line_data, tuple) for line_data in data):
-            raise ValueError("Matrix lines must be tuple")
-
-        if any(isinstance(item, Iterable) for line_data in data for item in
-               line_data):
-            raise ValueError("Matrix lines must be tuple")
+        if any(hasattr(item, '__len__') for line in data for item in line):
+            raise TypeError("Matrix item type not supported")
 
         if not all([len(line) == len(data) for line in data]):
             raise ValueError("Matrix must be even-sided")
 
     @classmethod
     def unity(cls, dimension: int):
-        """Generate a matrix with a main diagonal of 1's, everything else
-        is 0's.
-        """
-        return cls(*[tuple((0,) * j + (1,) + (0,) * i) for j, i in enumerate(
-            reversed(range(dimension)))])
+        """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
+        return cls(
+            tuple([tuple((0,) * i + (1,) + (0,) * (dimension - i - 1)) for i in
+                   range(dimension)]))
 
     @classmethod
     def ones(cls, dimension):
         """Generate a matrix full of 1's."""
-        return cls(*([tuple((1,) * dimension) for _ in range(dimension)]))
+        return cls(tuple(([tuple((1,) * dimension) for _ in
+                           range(dimension)])))
 
     def __repr__(self):
-        return f'<{self.__class__.__name__} object {hash(self)}>'
+        return f'{self.__class__.__name__}{self._matrix}'
 
     def __str__(self):
         """Return a string presentation of the matrix instance."""
@@ -89,7 +84,7 @@ class Matrix:
 
             calculated_matrix.append(tuple(calculated_line))
 
-        return Matrix(*calculated_matrix)
+        return Matrix(tuple(calculated_matrix))
 
     def _by_scalar(self, scalar, operation):
         """Calculate operation on matrix by scalar.
@@ -114,7 +109,7 @@ class Matrix:
 
             calculated_matrix.append(tuple(calculated_line))
 
-        return Matrix(*calculated_matrix)
+        return Matrix(tuple(calculated_matrix))
 
     def __add__(self, other):
         """Add two matrices."""
@@ -122,33 +117,24 @@ class Matrix:
 
     def __sub__(self, other):
         """Subtract two matrices."""
-        return self._matrix_by_matrix(other, operator.sub)
+        return self + (-1 * other)
 
     def __mul__(self, other):
         """Multiply two matrices or scalar by matrix."""
-        if is_number(other):
-            return self._by_scalar(other, operator.mul)
+        if isinstance(other, Matrix):
+            return self._matrix_by_matrix(other, operator.mul)
 
-        return self._matrix_by_matrix(other, operator.mul)
+        return self._by_scalar(other, operator.mul)
 
     def __rmul__(self, other):
         """Define to handle reverse multiplication (10 * matrix) as regular
         multiplication."""
-        return self.__mul__(other)
+        return self * other
 
     def __truediv__(self, other):
-        """Divide two matrices or scalar by matrix. Returns floats if
+        """Divide matrix by scalar. Returns floats if
         needed."""
-        if not is_number(other):
-            raise TypeError(
-                f'Matrices division not supported on {type(other)}')
-
-        return self._by_scalar(other, operator.truediv)
-
-    def __floordiv__(self, other):
-        """Define to handle floor division (5 / 2 = 2) as regular
-        division."""
-        return self.__truediv__(other)
+        return self * (1 / other)
 
     def _check_bool_matrix(self, matrix, operation):
         """Return true if all the items in the matrix are true else false."""
@@ -164,8 +150,7 @@ class Matrix:
     def __ne__(self, other):
         """Compare two matrices. Returns true if they are not equal else
         false."""
-        return self._check_bool_matrix(
-            self._matrix_by_matrix(other, operator.ne), operator.ne)
+        return not self == other
 
     def __iter__(self):
         """Enable to iterate on a matrix lines'."""
@@ -174,11 +159,3 @@ class Matrix:
     def __hash__(self):
         """Enable to use a matrix as key of a dictionary."""
         return hash(self.tuples)
-
-
-def is_number(s):
-    try:
-        float(s)
-        return True
-    except TypeError:
-        return False
--- a/matrix/matrix.py
+++ b/matrix/matrix.py
@@ -103,7 +103,7 @@ class Matrix:
         """
         return [tuple([operation(self.tuples[i][j], other.tuples[i][j]) for j
                        in range(len(self.tuples))]) for i in range(len(
-            self.tuples))]
+                        self.tuples))]
 
     def _calculate_scalar_on_matrix(self, other, operation):
         """Does mathematical operation on scalar and matrix.
@@ -146,8 +146,8 @@ class Matrix:
 
             return Matrix(*matrix_data)
 
-        elif any([isinstance(other, int) or isinstance(other, float)]) and any([
-            operation == operator.mul, operation == operator.truediv]):
+        elif any([isinstance(other, int) or isinstance(other, float)]) and any(
+                [operation == operator.mul, operation == operator.truediv]):
             return Matrix(*self._calculate_scalar_on_matrix(other, operation))
 
         else:
--- a/matrix/matrix.py
+++ b/matrix/matrix.py
@@ -1,3 +1,5 @@
+"""Matrix data type class."""
+
 import operator
 from collections.abc import Iterable
 
@@ -9,7 +11,7 @@ class Matrix:
         _matrix (tuple): contains the data about the matrix. Each line of
             the matrix is represented by a tuple.
     Note:
-        This implementation only supports even-side matrices (n * n).
+        This implementation only supports squared matrices (n * n).
     """
 
     def __init__(self, *lines):
@@ -35,7 +37,7 @@ class Matrix:
 
     @classmethod
     def unity(cls, dimension: int):
-        """Generates a matrix with a main diagonal of 1's, everything else
+        """Generate a matrix with a main diagonal of 1's, everything else
         is 0's.
         """
         return cls(*[tuple((0,) * j + (1,) + (0,) * i) for j, i in enumerate(
@@ -43,85 +45,64 @@ class Matrix:
 
     @classmethod
     def ones(cls, dimension):
-        """Generates a matrix full of 1's."""
+        """Generate a matrix full of 1's."""
         return cls(*([tuple((1,) * dimension) for _ in range(dimension)]))
 
     def __repr__(self):
-        return super(Matrix, self).__repr__()
+        return f'<{self.__class__.__name__} object {hash(self)}>'
 
     def __str__(self):
-        """Returns a string presentation of the matrix instance."""
+        """Return a string presentation of the matrix instance."""
         return repr(self._matrix)
 
     @property
     def tuples(self):
-        """Returns the matrix instance as a tuple of tuples."""
+        """Return the matrix instance as a tuple of tuples."""
         return self._matrix
 
-    def _by_matrix(self, operation, dimension_of_other,
-                   self_item, other_item):
-        """Calculates operation on two items.
+    def _matrix_by_matrix(self, other, operation):
+        """Calculate operation on two matrices.
 
         Args:
             operation (function): the function of the operator.
-            dimension_of_other (int): the number of lines in the other matrix.
-            self_item (number): current item from the current matrix instance.
-            other_item (number): current item from the other matrix instance.
+            other (Matrix): the other matrix to operate on.
 
         Returns:
-            number, the result of the operation.
+            Matrix: the result of the operation.
 
         Note:
-            Boolean operations supports different dimensions matrices but
-            mathematical operations don't (except matrix division).
+            Boolean operations supports un-squared matrices but
+            mathematical operations don't.
         """
-
-        if dimension_of_other != len(self.tuples) and not any(
+        if len(other.tuples) != len(self.tuples) and not any(
                 [operation == operator.eq, operation == operator.ne]):
             raise ValueError("Matrices length don't match")
 
-        if operation == operator.truediv:
-            raise ValueError(f"{operation} not supported on "
-                             f"{self.__class__.__name__}")
+        calculated_matrix = []
 
-        return operation(self_item, other_item)
+        for i in range(len(self.tuples)):
 
-    def _by_scalar(self, operation, item, scalar):
-        """Calculates operation on item by scalar.
+            calculated_line = []
+            for j in range(len(self.tuples)):
+                calculated_line.append(
+                    operation(self.tuples[i][j], other.tuples[i][j]))
+
+            calculated_matrix.append(tuple(calculated_line))
+
+        return Matrix(*calculated_matrix)
+
+    def _by_scalar(self, scalar, operation):
+        """Calculate operation on matrix by scalar.
 
         Args:
             operation (function): the function of the operator.
-            item (number): current item from the current matrix instance.
             scalar (number): the scalar to operate by.
 
         Returns:
-            number, the result of the operation.
+            Matrix: the result of the operation.
 
         Note:
             Scalar must be a number. Operations with scalars are only * and /.
-        """
-        if is_number(scalar) and not any([operation == operator.mul,
-                                          operation == operator.truediv]):
-            raise ValueError(f'{operation} not supported on {type(scalar)}')
-
-        return operation(item, scalar)
-
-    def _calculate_matrix_operation(self, other, operation):
-        """Does all the operations that can be done on a matrix.
-
-        Args:
-            other (Matrix or number): the other object to do the operation.
-            operation (function): the function of the desired operator.
-
-        Returns:
-            If boolean operation, returns true if operation returned true
-            else false. If mathematical operation, returns a new matrix with
-            the calculated data.
-
-        Note:
-            Supported operations are addition, subtraction, multiplication and
-            division(on matrix or by scalar). Also supports boolean
-            operators == and !=.
         """
         calculated_matrix = []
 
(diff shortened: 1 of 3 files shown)


matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+    implementation only supports even-side matrices (n * n).
+
+    Attributes:
+        _matrix (tuple): contains the data about the matrix. Each line of
+                         the matrix is represented by a tuple.
+    """
+    __slots__ = '_matrix'
+
+    def __init__(self, *lines):
+        if not self._is_valid_data(*lines):
```
avrahamshukron: "data is not valid" is not a good error message.

Instead of checking the return value of `_is_valid_data`, simply call it `validate` and let it throw the `ValueError` with the correct message.

Less code, more descriptive errors

matrix/tests_matrix.py
```diff
@@ -0,0 +1,190 @@
+import unittest
+
+from matrix import Matrix
+
+
+class TestMatrix(unittest.TestCase):
```
osherdp: You might notice this test hasn't been run by `pytest`
here's why:
https://docs.pytest.org/en/latest/unittest.html

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
+import operator
+from itertools import repeat
+
+
+class Matrix:
+    """An immutable object representing a matrix data-type. This
+    implementation only supports even-side matrices (n * n).
```
osherdp: The first sentence of a docstring should span over exactly one line

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+
+    Attributes:
+        _matrix (tuple): contains the data about the matrix. Each line of
+                         the matrix is represented by a tuple.
```
osherdp: Indentation here is wrong
see here: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

avrahamshukron (changes requested):

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    __slots__ = '_matrix'
+
+    def __init__(self, *lines):
+        if not self._is_valid_data(*lines):
+            raise ValueError('Data is not valid')
```
osherdp: You can and should raise exception with a more indicative message
Let the validity function raise the exception and set the appropriate message (e.g. `"Element {} is of type {}, expected a tuple"`)

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+
+        super(Matrix, self).__setattr__('_matrix', tuple(lines))
```
osherdp: Very unreadable
What's about `self._matrix = lines`?

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def __setattr__(self, key, value):
```
osherdp: This defense is not needed
If someone wants to change the value he can, for example, override this function

You just need to make it not too easy for him, e.g. using tuples instead of lists

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+        """Defines that Matrix attributes are immutable."""
+        raise AttributeError(f'{self.__class__} instance cannot be modified')
+
+    @staticmethod
+    def _is_valid_data(*data: tuple) -> bool:
+        """Checks if the data given is valid. Valid data is defined if it
```
osherdp: First word of docstring should be a declarative verb, e.g. `Check`

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+            'not equal', return true if one or more items are not equal to
+            their parallel item/s in other, else false.
+        """
+        if operation == operator.eq:
+            return all(
+                [item for line_data in data for item in line_data])
+
+        elif operation == operator.ne:
+            return any(
+                [item for line_data in data for item in line_data])
```
avrahamshukron: Instead of `if-elif` which is considered a code-smell in Python, you could have just called the correct function as follows:

```suggestion
        func = {operator.eq: all, operator.ne: any}.get(operation)
        return func(
                [item for line_data in data for item in line_data])
```

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+        if not self._is_valid_data(*lines):
+            raise ValueError('Data is not valid')
+
+        super(Matrix, self).__setattr__('_matrix', tuple(lines))
+
+    def __setattr__(self, key, value):
+        """Defines that Matrix attributes are immutable."""
+        raise AttributeError(f'{self.__class__} instance cannot be modified')
+
+    @staticmethod
```
osherdp: Usage of `staticmethod` is highly rare and is usually not recommended
If your IDE yells about the unused `self`, you can either ignore it or define this function outside of class's scope

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+    def _is_valid_data(*data: tuple) -> bool:
+        """Checks if the data given is valid. Valid data is defined if it
+        contains tuples of the same length (Matrix only supports n * n) and
+        their value is an integer."""
+        # check if each line is a tuple.
+        if not all([isinstance(line_data, tuple) for line_data in data]):
```
osherdp: you can remove the squared brackets. `all` and `any` can get generators as input as well as any iterable

matrix/tests_matrix.py
```diff
@@ -0,0 +1,190 @@
...
+    def _test_initialization(self, *data):
+        try:
+            Matrix(*data)
+        except ValueError:
+            self.fail("Initialization raised ValueError")
```
avrahamshukron: Exception raised unexpectedly will fail the test so you don't have to `try-expect` it

avrahamshukron (changes requested):

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+            return False
+
+        # check if each item in each line is an int.
+        if not all([any([isinstance(item, int), isinstance(item, float)]) for
```
osherdp: Checking types is not very Pythonic
If the user wants to put elements of complex numbers or even a class that behaves just like an `int`, he should be able to

It's also called duck-typing

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+    def __setattr__(self, key, value):
+        """Defines that Matrix attributes are immutable."""
+        raise AttributeError(f'{self.__class__} instance cannot be modified')
+
+    @staticmethod
+    def _is_valid_data(*data: tuple) -> bool:
+        """Checks if the data given is valid. Valid data is defined if it
+        contains tuples of the same length (Matrix only supports n * n) and
+        their value is an integer."""
+        # check if each line is a tuple.
```
osherdp: Comments in this function are redundant here as the code is pretty self-explanatory

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+            return False
+
+        # checks if the given data is of a n * n matrix.
+        if not all([len(line) == len(data) for line in data]):
+            return False
+
+        return True
+
+    @classmethod
+    def unity(cls, number_lines: int):
```
osherdp: rename `number_lines` -> `dimension`

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+            range(number_lines)]
+
+        return cls(*tuple(
+            [tuple(map(int, tuple_chars)) for tuple_chars in tuples_in_chars]))
+
+    @classmethod
+    def ones(cls, number_lines):
+        """Generates a matrix full of 1's."""
+        return cls(*tuple(
+            [tuple(repeat(1, number_lines)) for _ in range(number_lines)]))
```
osherdp: You can repeat the lines just like you used `repeat` to repeat the `1`
Alternatively, `(1,) * number_lines` works just the same

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+
+        return True
+
+    @classmethod
+    def unity(cls, number_lines: int):
+        """Generates a matrix with a main diagonal of 1's, everything else
+        is 0's.
+        """
+        tuples_in_chars = [
+            list(str('1' + ('0' * i)).zfill(number_lines)[::-1]) for i in
```
osherdp: That's rather cryptic
simplify this

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+
+    @classmethod
+    def ones(cls, number_lines):
+        """Generates a matrix full of 1's."""
+        return cls(*tuple(
+            [tuple(repeat(1, number_lines)) for _ in range(number_lines)]))
+
+    def __repr__(self):
+        """Returns a string presentation of the matrix instance."""
+        return f"({', '.join([str(line) for line in self._matrix])})"
```
osherdp: `repr` should perform: `eval(repr(obj)) == obj`

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+            If the operation is 'equal', returns true if all items are equal
+            to their parallel items in other, else false. If operation is
+            'not equal', return true if one or more items are not equal to
+            their parallel item/s in other, else false.
+        """
+        if operation == operator.eq:
+            return all(
+                [item for line_data in data for item in line_data])
+
+        elif operation == operator.ne:
```
osherdp: No need in this...
You should implement "equal" just like you did, and then the implementation of "not-equal" is pretty straight-forward afterward ;)

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        return [tuple([operation(self.tuples[i][j], other.tuples[i][j]) for j
+                       in range(len(self.tuples))]) for i in range(len(
+                        self.tuples))]
+
+    def _calculate_scalar_on_matrix(self, other, operation):
+        """Does mathematical operation on scalar and matrix.
+
+        Args:
+            other (int or float): the scalar.
```
osherdp: `int or float` -> `number`

matrix/tests_matrix.py
```diff
@@ -0,0 +1,190 @@
...
+
+        for line in Matrix.ones(2):
+            self.assertEqual(line, (1, 1))
+
+    def test_hash(self):
+        try:
+            d = {Matrix((1, 1), (2, 2)): 2, Matrix((1, 1), (2, 3)): 3}
+
+        except TypeError:
+            self.fail('unhashable type')
```
osherdp: no need. An exception will make the test fail nonetheless

matrix/tests_matrix.py
```diff
@@ -0,0 +1,190 @@
...
+
+        else:
```
osherdp: As a rule of thumb:
`if`/`else` almost never should be used in tests
`for` should be rare as well

matrix/tests_matrix.py
```diff
@@ -0,0 +1,190 @@
...
+            _ = m1 / m3
+
+        with self.assertRaises(ValueError):
+            _ = m1 / "hi"
+
+        with self.assertRaises(ValueError):
+            _ = m1 / [1, 2, 3]
+
+        with self.assertRaises(ValueError):
+            _ = m1 / ["hi", "i'm", "a test"]
```
osherdp: Lots of those tests are redundant and have no actual value
You don't need to check your function against exotic types

matrix/tests_matrix.py
```diff
@@ -0,0 +1,190 @@
...
+
+        with self.assertRaises(ValueError):
+            _ = m1 / (1, 2)
+
+        with self.assertRaises(ValueError):
+            _ = m1 / ("hi", "i'm", "a test")
+
+    def test_equal(self):
+        self.assertTrue(
+            Matrix.unity(3) == Matrix((1, 0, 0), (0, 1, 0), (0, 0, 1)))
```
osherdp: Use `self.assertEqual`
here and everywhere

matrix/tests_matrix.py
```diff
@@ -0,0 +1,190 @@
...
+        self.assertTrue(
+            Matrix.unity(3) != Matrix.ones(3))
+
+        self.assertTrue(
+            Matrix.unity(3) != Matrix.ones(4))
+
+        self.assertFalse(Matrix((1, 2), (3, 4)) != Matrix((1, 2), (3, 4)))
+
+    def test_iteration(self):
+        self.assertNotEqual(list(iter(Matrix.ones(3))), [])
```
osherdp: redundant

matrix/tests_matrix.py
```diff
@@ -0,0 +1,190 @@
...
+
+        for line in Matrix.ones(2):
+            self.assertEqual(line, (1, 1))
```
osherdp: just `self.assertEqual(tuple(Matrix.ones(2)), ((1, 1), (1, 1))`

matrix/matrix.py
```diff
@@ -0,0 +1,189 @@
+import operator
+from itertools import repeat
+
+
+class Matrix:
+    """An immutable object representing a matrix data-type. This
+    implementation only supports even-side matrices (n * n).
```
rfire01: First row in doc should have short explanation about it.
It should be exactly 1 row (short and descriptive).
If you make something complex its ok to add notes in the doc below to explain it

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+            other (int or float): the scalar.
+            operation (function): the mathematical function to operate.
+
+        Returns:
+            List of tuples which contains the calculated matrix data.
+        """
+        return [tuple([operation(self.tuples[i][j], other) for j in range(
+            len(self.tuples))]) for i in range(len(self.tuples))]
+
+    def _operations_on_matrix(self, other, operation):
```
osherdp: This method tries to do too much
See how you can reuse actions and be more concrete with every operator

matrix/matrix.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __setattr__(self, key, value):
+        """Defines that Matrix attributes are immutable."""
+        raise AttributeError(f'{self.__class__} instance cannot be modified')
+
+    @staticmethod
+    def _validate(*data: tuple):
+        """Checks if the data given is valid. Valid data is defined if it
+        contains tuples of the same length (Matrix only supports n * n) and
+        their value is an integer."""
```
rfire01: same

matrix/matrix.py
```diff
@@ -0,0 +1,198 @@
...
+
+        Args:
+            other (int or float): the scalar.
+            operation (function): the mathematical function to operate.
+
+        Returns:
+            List of tuples which contains the calculated matrix data.
+        """
+        return [tuple([operation(self.tuples[i][j], other) for j in range(
+            len(self.tuples))]) for i in range(len(self.tuples))]
```
osherdp: It would be better if you format it to multiple lines with a clear structure of what's going on here...

osherdp (changes requested):

osherdp:
First of all, not bad for the first submission
Don't be intimidated by the number of comments, most of them are pretty technical

matrix/matrix.py
```diff
@@ -0,0 +1,189 @@
...
+
+        Args:
+            operation (function): the boolean function to operate.
+            data (list): the data of the matrix (list of tuples).
+
+        Returns:
+            If the operation is 'equal', returns true if all items are equal
+            to their parallel items in other, else false. If operation is
+            'not equal', return true if one or more items are not equal to
+            their parallel item/s in other, else false.
```
rfire01: first you should write the type that returned.
You can write it in much simpler way (for example, check if matrices equal element-wise)

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
+import unittest
+
+from matrix import Matrix
+
+
+class TestMatrix(unittest.TestCase):
+    def test_initialization(self):
+        Matrix((1, 2), (3, 4))
+        Matrix((1, 2), (5.5, 4))
```
rfire01: ???

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
+import unittest
+
```
rfire01: you don't have docs

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+        with self.assertRaises(ValueError):
+            Matrix((1, 'a'), (3, 4))
+
+        with self.assertRaises(ValueError):
+            Matrix((1, 2), (3, [4, 5, 6]))
+
+        with self.assertRaises(ValueError):
+            Matrix((1, 2), (3, 4, 5))
+
+    def test_modification(self):
```
rfire01: what about changing a single element?

matrix/matrix.py
yonatanevantal: Fixed in last commit.

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+        with self.assertRaises(AttributeError):
+            m._matrix = ((1, 2), (3, 4))
+
+        with self.assertRaises(AttributeError):
+            m._matrix = "hello world!"
+
+    def test_str(self):
+        m = Matrix((1, 1, 1), (2, 2, 2), (3, 3, 3))
+
+        self.assertEqual(str(m), str(((1, 1, 1), (2, 2, 2), (3, 3, 3))))
```
rfire01: Why not save the tuple in variable first?
This way you don't need to duplicate it, and its easy to change if needed

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def test_tuples(self):
+        m = Matrix((1, 1), (2, 2))
+
+        self.assertEqual(m.tuples, ((1, 1), (2, 2)))
```
rfire01: same

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+        self.assertEqual(
+            Matrix.ones(2), Matrix((1, 1), (1, 1)))
+
+        self.assertEqual(
+            Matrix.ones(3), Matrix((1, 1, 1), (1, 1, 1), (1, 1, 1)))
+
+    def test_add(self):
+        m1 = Matrix.ones(3)
+        m2 = Matrix.ones(3)
+        m3 = Matrix.ones(2)
```
rfire01: why not use more interesting matrix? at least one of them

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+            _ = m1 + m3
+
+        with self.assertRaises(ValueError):
+            _ = m1 + 10
+
+        with self.assertRaises(ValueError):
+            _ = m1 + 0.5
+
+        with self.assertRaises(ValueError):
+            _ = m1 - "0.5"
```
rfire01: if you don't save the value, you need to use "="

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+            _ = m1 - m3
+
+        with self.assertRaises(ValueError):
+            _ = m1 - 10
+
+        with self.assertRaises(ValueError):
+            _ = m1 - 0.5
+
+        with self.assertRaises(ValueError):
+            _ = m1 - "0.5"
```
rfire01: same

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+            _ = m1 * [1, 2, 3]
+
+        with self.assertRaises(ValueError):
+            _ = m1 * ["hi", "i'm", "a test"]
+
+        with self.assertRaises(ValueError):
+            _ = m1 * (1, 2)
+
+        with self.assertRaises(ValueError):
+            _ = m1 * ("hi", "i'm", "a test")
```
rfire01: same

matrix/tests_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def test_div(self):
+        m1 = Matrix((1, 2), (3, 4))
+        m2 = Matrix((5, 6), (7, 8))
+        m3 = Matrix.ones(4)
+
+        self.assertEqual(m1 / m2, Matrix((0.2, 2 / 6), (3 / 7, 0.5)))
```
rfire01: This should not be supported, its not a legal matrix operation

matrix/matrix.py
yonatanevantal: I don't understand... I wrote 'Checks'.

rfire01: Checks means he does something
Check means you are told to check
Think about it as the first word should be a command

matrix/matrix.py
yonatanevantal: Fixed in last commit.

matrix/tests_matrix.py
yonatanevantal: Fixed in previous commit.

matrix/tests_matrix.py
yonatanevantal: So there is no need to seperate the tests of == and !=, right?

osherdp: It's actually a good idea to separate them, as they are different methods

Each small change in the code should preferably break exactly one test. It's not always possible, but it's preferable

matrix/matrix.py
yonatanevantal: But what about the case of items that are iterable? I think the definition of the exercise does not say to support that case, so it must be blocked.

matrix/tests_matrix.py
yonatanevantal: Got it, thanks.

matrix/tests_matrix.py
yonatanevantal: Can't figure it out. Maybe live CR on this one.

matrix/matrix.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+class Matrix:
+    """An immutable object representing a matrix data-type.
+
+    Attributes:
+        _matrix (tuple): contains the data about the matrix. Each line of
+            the matrix is represented by a tuple.
+    Note:
+        This implementation only supports even-side matrices (n * n).
```
rfire01: its called  squared matrices

matrix/matrix.py
```diff
@@ -0,0 +1,202 @@
...
+        return cls(*[tuple((0,) * j + (1,) + (0,) * i) for j, i in enumerate(
+            range(dimension - 1, -1, -1))])
+
+    @classmethod
+    def ones(cls, dimension):
+        """Generates a matrix full of 1's."""
+        return cls(*([tuple((1,) * dimension) for _ in range(dimension)]))
+
+    def __repr__(self):
+        return super(Matrix, self).__repr__()
```
rfire01: the purpose here, is not that you will use already existing repr, but that you will create one yourself

matrix/matrix.py
```diff
@@ -0,0 +1,202 @@
...
+        return repr(self._matrix)
+
+    @property
+    def tuples(self):
+        """Returns the matrix instance as a tuple of tuples."""
+        return self._matrix
+
+    def _by_matrix(self, operation, dimension_of_other,
+                   self_item, other_item):
+        """Calculates operation on two items.
```
rfire01: first word should be imperative verb (fix in entire code)

matrix/matrix.py
```diff
@@ -0,0 +1,202 @@
...
+
+        Args:
+            operation (function): the function of the operator.
+            dimension_of_other (int): the number of lines in the other matrix.
```
rfire01: why do you need to get this? you can easily calculate it

matrix/matrix.py
```diff
@@ -0,0 +1,202 @@
...
+    def __str__(self):
+        """Returns a string presentation of the matrix instance."""
+        return repr(self._matrix)
+
+    @property
+    def tuples(self):
+        """Returns the matrix instance as a tuple of tuples."""
+        return self._matrix
+
+    def _by_matrix(self, operation, dimension_of_other,
```
rfire01: this function name is not informative

matrix/matrix.py
```diff
@@ -0,0 +1,202 @@
...
+            other_item (number): current item from the other matrix instance.
+
+        Returns:
+            number, the result of the operation.
+
+        Note:
+            Boolean operations supports different dimensions matrices but
+            mathematical operations don't (except matrix division).
+        """
+
```
rfire01, line 84: remove

matrix/matrix.py
```diff
@@ -0,0 +1,202 @@
...
+        Note:
+            Scalar must be a number. Operations with scalars are only * and /.
+        """
+        if is_number(scalar) and not any([operation == operator.mul,
+                                          operation == operator.truediv]):
+            raise ValueError(f'{operation} not supported on {type(scalar)}')
+
+        return operation(item, scalar)
+
+    def _calculate_matrix_operation(self, other, operation):
```
rfire01: usually when writing code there is a trade-off
Generic code vs readable code.
Creating one function to rule them all (function that does everything), might be good as it support any action, but makes the code much less readable.

You can see that in your code, by the fact that everytime you call it, you must run validation on all operations, but the validation is not always valid for all operations, so you must add another if for it, etc.
This makes your code look clumsy

matrix/tests_matrix.py
```diff
@@ -0,0 +1,147 @@
+import unittest
```
rfire01: File should always start with a doc line that describes the file

matrix/matrix.py
```diff
@@ -0,0 +1,202 @@
+import operator
```
rfire01: File should always start with a doc line that describes the file

matrix/tests_matrix.py
rfire01: There is no doc

matrix/tests_matrix.py
```diff
@@ -0,0 +1,147 @@
...
+from matrix import Matrix
+
+
+@pytest.fixture(scope="module")
+def db_class(request):
+    # set a class attribute on the invoking test context
+    request.cls.db = Matrix()
+
+
+class TestMatrix(unittest.TestCase):
```
rfire01: Look at the purpose of unitest \ testcase \ testsuite
Sometimes its better to separate the test to several cases (for example, 1 case for testing creating function, 1 for operations, etc.)

rfire01 (changes requested):

matrix/matrix.py
yonatanevantal: because I don't get the entire other matrix, only the current item.

matrix/tests_matrix.py
yonatanevantal: Do you mean to seperate to different classes?

rfire01: Yes

matrix/matrix.py
```diff
@@ -0,0 +1,184 @@
...
+        return cls(*[tuple((0,) * j + (1,) + (0,) * i) for j, i in enumerate(
+            range(dimension - 1, -1, -1))])
+
+    @classmethod
+    def ones(cls, dimension):
+        """Generate a matrix full of 1's."""
+        return cls(*([tuple((1,) * dimension) for _ in range(dimension)]))
+
+    def __repr__(self):
+        return f'<{self.__class__.__name__} object {hash(self)}>'
```
osherdp: Dude, read that exercise. It shouldn't look like this

matrix/matrix.py
```diff
@@ -0,0 +1,184 @@
...
+        """Add two matrices."""
+        return self._matrix_by_matrix(other, operator.add)
+
+    def __sub__(self, other):
+        """Subtract two matrices."""
+        return self._matrix_by_matrix(other, operator.sub)
+
+    def __mul__(self, other):
+        """Multiply two matrices or scalar by matrix."""
+        if is_number(other):
```
osherdp: Delete the is_number usage

Check if it's an instance of Matrix, and perform the scalar multiplication. If it's not, assume it's a scalar

matrix/matrix.py
```diff
@@ -0,0 +1,184 @@
...
+        """Check if the data given is valid.
+
+        Note:
+            Valid data is defined if it contains tuples of the same length
+            (Matrix only supports n * n), items can't be iterable.
+        """
+        if not all(isinstance(line_data, tuple) for line_data in data):
+            raise ValueError("Matrix lines must be tuple")
+
+        if any(isinstance(item, Iterable) for line_data in data for item in
```
osherdp: Remove all isinstance usage, the Pythonic way is to duck-type everything

The only place you must use isinstance is the multiplication

matrix/matrix.py
```diff
@@ -0,0 +1,184 @@
...
+               line_data):
+            raise ValueError("Matrix lines must be tuple")
+
+        if not all([len(line) == len(data) for line in data]):
+            raise ValueError("Matrix must be even-sided")
+
+    @classmethod
+    def unity(cls, dimension: int):
+        """Generate a matrix with a main diagonal of 1's, everything else
```
osherdp: Please, make it exactly one line, not two
Here and everywhere

matrix/matrix.py
```diff
@@ -0,0 +1,184 @@
...
+    """An immutable object representing a matrix data-type.
+
+    Attributes:
+        _matrix (tuple): contains the data about the matrix. Each line of
+            the matrix is represented by a tuple.
+    Note:
+        This implementation only supports squared matrices (n * n).
+    """
+
+    def __init__(self, *lines):
```
osherdp: In this case, I don't like the usage of *args
Just get the tuple as an argument

matrix/matrix.py
```diff
@@ -0,0 +1,184 @@
...
+
+        if not all([len(line) == len(data) for line in data]):
+            raise ValueError("Matrix must be even-sided")
+
+    @classmethod
+    def unity(cls, dimension: int):
+        """Generate a matrix with a main diagonal of 1's, everything else
+        is 0's.
+        """
+        return cls(*[tuple((0,) * j + (1,) + (0,) * i) for j, i in enumerate(
```
osherdp: Pay attention that i+j+1==dimension
So you can delete either i or j

matrix/matrix.py
```diff
@@ -0,0 +1,184 @@
...
+    def __truediv__(self, other):
+        """Divide two matrices or scalar by matrix. Returns floats if
+        needed."""
+        if not is_number(other):
+            raise TypeError(
+                f'Matrices division not supported on {type(other)}')
+
+        return self._by_scalar(other, operator.truediv)
+
+    def __floordiv__(self, other):
```
osherdp: Floordir should either be deleted or perform the same as the default floor division, not being exactly like the regular div

matrix/matrix.py
```diff
@@ -0,0 +1,184 @@
...
+        """Multiply two matrices or scalar by matrix."""
+        if is_number(other):
+            return self._by_scalar(other, operator.mul)
+
+        return self._matrix_by_matrix(other, operator.mul)
+
+    def __rmul__(self, other):
+        """Define to handle reverse multiplication (10 * matrix) as regular
+        multiplication."""
+        return self.__mul__(other)
```
osherdp: Replace by self * other

matrix/matrix.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def __eq__(self, other):
+        """Compare two matrices. Returns true if they are equal else false."""
+        return self._check_bool_matrix(
+            self._matrix_by_matrix(other, operator.eq), operator.eq)
+
+    def __ne__(self, other):
+        """Compare two matrices. Returns true if they are not equal else
+        false."""
+        return self._check_bool_matrix(
```
osherdp: Just `not self == other`
Find other places where you can reuse operators, e.g. the subtraction operator

osherdp (changes requested):

matrix/test_unittest_matrix.py
```diff
@@ -0,0 +1,177 @@
...
+    def test_tuples(self):
+        """Test tuple which should return the matrix data as tuple."""
+        data = ((1, 1), (2, 2))
+        m = Matrix(data)
+
+        self.assertEqual(m.tuples, data)
+
+    def test_unity(self):
+        """Test unity() which should return a matrix with 1's in the main
+        diagonal and 0's elsewhere.
```
rfire01: the first line of the doc should be single line

matrix/test_unittest_matrix.py
```diff
@@ -0,0 +1,177 @@
...
+    def test_hash(self):
+        """Test __hash__ function and uses dictionary keys."""
+        d = {Matrix(((1, 1), (2, 2))): 2, Matrix(((1, 1), (2, 3))): 3}
+
+        self.assertEqual(d, {Matrix(((1, 1), (2, 2))): 2,
+                             Matrix(((1, 1), (2, 3))): 3})
+
+        self.assertNotEqual(hash(Matrix.ones(2)), hash(Matrix.unity(2)))
+
+
```
rfire01, line 172: what about testsuite?

matrix/matrix.py
```diff
@@ -0,0 +1,161 @@
...
+    def unity(cls, dimension: int):
+        """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
+        return cls(
+            tuple([tuple((0,) * i + (1,) + (0,) * (dimension - i - 1)) for i in
+                   range(dimension)]))
+
+    @classmethod
+    def ones(cls, dimension):
+        """Generate a matrix full of 1's."""
+        return cls(tuple(([tuple((1,) * dimension) for _ in
```
osherdp: `tuple((1,) * dimension)` is redundant, it's already a tuple
Also, you already have a shorter way to repeat objects (` * dimension`), why don't you use it?

matrix/matrix.py
```diff
@@ -0,0 +1,161 @@
...
+
+        if not all([len(line) == len(data) for line in data]):
+            raise ValueError("Matrix must be even-sided")
+
+    @classmethod
+    def unity(cls, dimension: int):
+        """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
+        return cls(
+            tuple([tuple((0,) * i + (1,) + (0,) * (dimension - i - 1)) for i in
+                   range(dimension)]))
```
osherdp:
```suggestion
            tuple([tuple((0,) * i + (1,) + (0,) * (dimension - i - 1))
                  for i in range(dimension)]))
```

matrix/matrix.py
```diff
@@ -0,0 +1,161 @@
...
+            raise TypeError("Matrix item type not supported")
+
+        if not all([len(line) == len(data) for line in data]):
+            raise ValueError("Matrix must be even-sided")
+
+    @classmethod
+    def unity(cls, dimension: int):
+        """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
+        return cls(
+            tuple([tuple((0,) * i + (1,) + (0,) * (dimension - i - 1)) for i in
```
osherdp: remove all unneeded list creation, here and everywhere

matrix/matrix.py
```diff
@@ -0,0 +1,161 @@
...
+                   range(dimension)]))
+
+    @classmethod
+    def ones(cls, dimension):
+        """Generate a matrix full of 1's."""
+        return cls(tuple(([tuple((1,) * dimension) for _ in
+                           range(dimension)])))
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}{self._matrix}'
```
osherdp: make sure `eval(repr(obj)) == obj`. This is the requirement for the `repr` method

matrix/matrix.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def _check_bool_matrix(self, matrix, operation):
+        """Return true if all the items in the matrix are true else false."""
+        func = {operator.eq: all, operator.ne: any}.get(operation)
+        return func(
+            [item for line_data in matrix.tuples for item in line_data])
+
+    def __eq__(self, other):
+        """Compare two matrices. Returns true if they are equal else false."""
+        return self._check_bool_matrix(
```
osherdp: you're overcomplicating it, just compare the tuples
then delete methods like `_check_bool_matrix` and `_by_scalar` as they have exactly one usage

matrix/test_unittest_matrix.py
```diff
@@ -0,0 +1,202 @@
...
+"""
+
+import unittest
+
+from matrix import Matrix
+
+
+class TestMatrixCreate(unittest.TestCase):
+    """Test the creation of a matrix and that an instance can't be modified.
+    """
```
osherdp: It's not followed by the PEP257 conventions. Please reread

matrix/matrix.py
```diff
@@ -0,0 +1,161 @@
...
+        self._matrix = lines
+
+    def _validate(self, data: tuple):
+        """Check if the data given is valid.
+
+        Note:
+            Valid data is defined if it contains tuples of the same length
+            (Matrix only supports n * n), items can't be iterable.
+        """
+        if any(hasattr(item, '__len__') for line in data for item in line):
```
osherdp: Also not Pythonic
Python will raise an exception if the object doesn't have `__len__`, so you don't need to implement anything by yourself

matrix/matrix.py
```diff
@@ -0,0 +1,161 @@
...
+        return self._matrix_by_matrix(other, operator.add)
+
+    def __sub__(self, other):
+        """Subtract two matrices."""
+        return self + (-1 * other)
+
+    def __mul__(self, other):
+        """Multiply two matrices or scalar by matrix."""
+        if isinstance(other, Matrix):
+            return self._matrix_by_matrix(other, operator.mul)
```
osherdp: It doesn't look like matrix multiplication
Are you sure about your implementation?

matrix/test_unittest_matrix.py
osherdp: @rfire01 I'm not sure testsuites are really interesting
No one creates them anymore, all test-engines can discover all tests anyway

osherdp (changes requested):

matrix/matrix.py
yonatanevantal: What do you mean? It passed the tests.

osherdp: The tests also don't look right
Do you know how matrix multiplication works?

yonatanevantal: Got the mistake, on it :)

matrix/test_unittest_matrix.py
yonatanevantal: So should I implement it?

matrix/matrix.py
yonatanevantal: So ignore the case of an iterable as a matrix item?

matrix/matrix.py
yonatanevantal: In the exercise description it seems that repr() just returns the class name and attribute.
`repr(a) == 'Matrix((1, 2), (3, 4))'`

matrix/test_unittest_matrix.py
rfire01: just understand what is it, no need to implement

matrix/matrix.py
osherdp: Yeah, and it doesn't seem like the code you wrote returns that
Look closely

matrix/matrix.py
osherdp: You anyway iterate over it, so if it's not iterable some exception will be raised

osherdp: yes

matrix/matrix.py
osherdp: Alright, there were typos all over the exercise
anyway, make sure that `Matrix(((1, 2), (3, 4))) == eval(repr(Matrix(((1, 2), (3, 4))))`. This is the most exact definition

osherdp (changes requested):

matrix/matrix.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _validate(self, data: tuple):
+        """Check if the data given is valid.
+
+        note:
+            Valid data is defined if it contains tuples of the same length
+            (Matrix only supports n * n), items can't be iterable.
+        """
+        # if any(hasattr(item, '__len__') for line in data for item in line):
+        #     raise TypeError("Matrix item type not supported")
```
rfire01: ?

matrix/matrix.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, lines):
+        self._validate(lines)
+        self._matrix = lines
+
+    def _validate(self, data: tuple):
+        """Check if the data given is valid.
+
+        note:
+            Valid data is defined if it contains tuples of the same length
+            (Matrix only supports n * n), items can't be iterable.
```
rfire01: why can't be iterable? it contain tuples, which are iterable

matrix/matrix.py
```diff
@@ -0,0 +1,146 @@
...
+        #     raise TypeError("Matrix item type not supported")
+
+        if not all([len(line) == len(data) for line in data]):
+            raise ValueError("Matrix must be even-sided")
+
+    @classmethod
+    def unity(cls, dimension: int):
+        """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
+        return cls(
+            tuple(tuple((0,) * i + (1,) + (0,) * (dimension - i - 1))
```
rfire01: no need for both tuples

matrix/matrix.py
```diff
@@ -0,0 +1,146 @@
...
+                  for i in range(dimension)))
```
rfire01: this doesn't look good.
Its not bad to split it to multiple lines (logic lines)

matrix/matrix.py
```diff
@@ -0,0 +1,146 @@
...
+        Returns:
+            Matrix: the result of the addition.
+        """
+        dimension = len(self.tuples)
+
+        if len(other.tuples) != dimension:
+            raise ValueError("Matrices length don't match")
+
+        calculated_data = [[0 for i in range(dimension)]
+                           for j in range(dimension)]
```
rfire01: if you must do it, then there is prettier way to do this

matrix/matrix.py
```diff
@@ -0,0 +1,146 @@
...
+
+        for i in range(len(self.tuples)):
+            for j in range(len(self.tuples)):
+                calculated_data[i][j] = self.tuples[i][j] + other.tuples[i][j]
+
+        calculated_data = [tuple(line) for line in calculated_data]
+        return Matrix(tuple(calculated_data))
```
rfire01: why not make it a tuple from the start?

matrix/matrix.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _matrix_mul_matrix(self, other):
```
rfire01, line 92: You do the exact same validation for multiple function... try to read about decorators

matrix/matrix.py
```diff
@@ -0,0 +1,146 @@
...
+        """Divide matrix by scalar. Return floats if needed."""
+        return self * (1.0 / other)
+
+    def __eq__(self, other):
+        """Compare two matrices. Return true if they are equal else false."""
+        return self.tuples == other.tuples
+
+    def __ne__(self, other):
+        """Compare two matrices. Return true if they are not equal else
+        false."""
```
rfire01: doc should be 1 line

matrix/matrix.py
```diff
@@ -0,0 +1,146 @@
...
+        return not self == other
+
+    def __iter__(self):
+        """Enable to iterate on a matrix lines'."""
+        return iter(self.tuples)
```
rfire01, line 146: tuples are already iterable

matrix/test_unittest_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+"""
+
+import unittest
+
+from matrix import Matrix
+
+
+class TestMatrixCreate(unittest.TestCase):
+    """Test the creation of a matrix and that an instance can't be modified."""
+
```
rfire01: remove

matrix/test_unittest_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+    def test_initialization(self):
+        """Test init function of Matrix."""
+        self.assertIsInstance(Matrix(((1, 2), (3, 4))), Matrix)
+        self.assertIsInstance(Matrix(((1, 2), (5.5, 4))), Matrix)
```
rfire01: This test is weird.
You create a new instance of Matrix, its very unlikely to return an instance that isn't matrix (you need to make effort to cause it).
What can happen, is the initialization fails of one of the reasons:
* You get exception in the init
* The instance doesn't represent what you wanted

matrix/test_unittest_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+        """Test that Matrix attribute can't be modified."""
+        m = Matrix(((1, 1), (2, 2)))
+
+        with self.assertRaises(TypeError):
+            m._matrix[0] = (1, 2)
+
+
+class TestMatrixFunctions(unittest.TestCase):
+    """Test the built-in functions of Matrix."""
+
```
rfire01: remove (in all classes)

matrix/test_unittest_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+    def test_tuples(self):
+        """Test tuple which should return the matrix data as tuple."""
+        data = ((1, 1), (2, 2))
+        m = Matrix(data)
+
+        self.assertEqual(m.tuples, data)
+
+    def test_unity(self):
+        """Test unity().
```
rfire01: This isn't a good doc (you put the doc in the note instead of here)

matrix/test_unittest_matrix.py
```diff
@@ -0,0 +1,181 @@
...
+        m1 = Matrix(((1, 2), (3, 4)))
+        m2 = Matrix(((5, 6), (7, 8)))
+        m3 = Matrix.ones(4)
+
+        self.assertEqual(m1 * m2, Matrix(((19, 22), (43, 50))))
+        self.assertEqual(m1 * 10, Matrix(((10, 20), (30, 40))))
+        self.assertEqual(m1 * 0.5, Matrix(((0.5, 1), (1.5, 2))))
+
+        # reverse mul
+        self.assertEqual(10 * m1, Matrix(((10, 20), (30, 40))))
```
rfire01: why don't you test m2*m1?

rfire01 (changes requested):

matrix/matrix.py
yonatanevantal: Incorrect documentation.

matrix/matrix.py
yonatanevantal: If I won't put in `iter()` it will raise: `TypeError: iter() returned non-iterator of type 'tuple'`

matrix/test_unittest_matrix.py
yonatanevantal: I didn't understand that, can you explain it in different words?

matrix/matrix.py
rfire01: you are right, my mistake.
__iter__ expects iterator, not iterable

matrix/test_unittest_matrix.py
rfire01: If you create an object of type A, then if the init didn't failed, it will always return an object of type A, so checking it is meaningless
If you are checking that constructor works then there are two things you need to check:
The first is that it doesn't fail, e.g. you divide by zero in the  constructor, so it will throw an exception and fail
The second, it doesn't do what you want. For example, creating unity matrix of dim=1 need to create (1,) or ((1,)) (depend on how you implemented it), but it gave instead 1 (without the tuple)

yonatanevantal: So I need to check if the constructor doesn't fail and that the return value is what I meant it would be?

rfire01: If you write a test for it, yes

matrix/matrix.py
```diff
@@ -1,6 +1,19 @@
 """Matrix data type class."""
 
 
+def _operation_method(func):
```
rfire01: doc

rfire01: give it meaningful name, what it does

matrix/matrix.py
```diff
@@ -19,21 +32,21 @@ def _validate(self, data: tuple):
...
     @classmethod
     def unity(cls, dimension: int):
         """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
-        return cls(
-            tuple(tuple((0,) * i + (1,) + (0,) * (dimension - i - 1))
-                  for i in range(dimension)))
+        unity_matrix = []
+
+        for i in range(dimension):
+            unity_matrix.append((0,) * i + (1,) + (0,) * (dimension - i - 1))
```
rfire01: No need to save it to list and then convert to tuple, save it from the start to tuple

matrix/matrix.py
```diff
@@ -62,20 +76,15 @@ def _add_matrix(self, other):
             Matrix: the result of the addition.
         """
         dimension = len(self.tuples)
+        empty_data = tuple([] for i in range(dimension))
```
rfire01, line 83: Actually all I meant here, that instead of
[0 i range(x)]
you can write
[0] * x

matrix/matrix.py
```diff
@@ -133,8 +134,7 @@ def __eq__(self, other):
         return self.tuples == other.tuples
 
     def __ne__(self, other):
-        """Compare two matrices. Return true if they are not equal else
-        false."""
+        """Compare two matrices. The opposite of equal."""
```
rfire01: doc should be a single sentence

matrix/test_unittest_matrix.py
```diff
@@ -11,12 +11,12 @@
 
 
 class TestMatrixCreate(unittest.TestCase):
-    """Test the creation of a matrix and that an instance can't be modified."""
-
     def test_initialization(self):
         """Test init function of Matrix."""
-        self.assertIsInstance(Matrix(((1, 2), (3, 4))), Matrix)
-        self.assertIsInstance(Matrix(((1, 2), (5.5, 4))), Matrix)
+        matrix_data = (19, 48), (19, 67)
```
rfire01: when using tuples, either write:
matrix = ((1,2),(3,4))
or:
a, b = (1,2), (3,4)
i.e. if you want a single variable, make it clear that it represent a single variable

matrix/test_unittest_matrix.py
```diff
@@ -30,14 +30,11 @@ def test_modification(self):
 
 
 class TestMatrixFunctions(unittest.TestCase):
```
rfire01, line 33: why you removed the doc?

matrix/test_unittest_matrix.py
```diff
@@ -30,14 +30,11 @@ def test_modification(self):
...
-    """Test the built-in functions of Matrix."""
-
     def test_representation(self):
         """Test __str__  and __repr__ functions."""
         data = ((1, 1, 1), (2, 2, 2), (3, 3, 3))
         m = Matrix(data)
 
-        self.assertEqual(str(m), str(data))
```
rfire01: why removed?

matrix/test_unittest_matrix.py
```diff
@@ -11,12 +11,12 @@
...
-    """Test the creation of a matrix and that an instance can't be modified."""
-
     def test_initialization(self):
         """Test init function of Matrix."""
-        self.assertIsInstance(Matrix(((1, 2), (3, 4))), Matrix)
-        self.assertIsInstance(Matrix(((1, 2), (5.5, 4))), Matrix)
+        matrix_data = (19, 48), (19, 67)
+        m = Matrix(matrix_data)
+
+        self.assertEqual(str(m), str(matrix_data))
```
rfire01: you got a different test to check str later.
If you only want to chechk the object was created correctly, why not compare m.data with matrix_data?

matrix/test_unittest_matrix.py
```diff
@@ -70,8 +62,6 @@ def test_ones(self):
 
 
 class TestMatrixOperators(unittest.TestCase):
```
rfire01, line 67: why removed doc?

matrix/test_unittest_matrix.py
```diff
@@ -116,10 +106,12 @@ def test_mul(self):
         self.assertEqual(m1 * m2, Matrix(((19, 22), (43, 50))))
         self.assertEqual(m1 * 10, Matrix(((10, 20), (30, 40))))
         self.assertEqual(m1 * 0.5, Matrix(((0.5, 1), (1.5, 2))))
+        # doesnt return the same as m1 * m2
```
rfire01: redundant

matrix/test_unittest_matrix.py
```diff
@@ -157,16 +149,12 @@ def test_not_equal(self):
 
 
 class TestMatrixIteration(unittest.TestCase):
```
rfire01, line 154: why you removed the classes docs?

yonatanevantal: Thought you meant to remove them in a previous comment. Should I leave the docs of the methods as they are? The names are pretty self-explanatory.

matrix/test_unittest_matrix.py
yonatanevantal: Just a mistake.

matrix/matrix.py
yonatanevantal: what I need is a tuple of lists. Using the multiplication option causes reference errors (I tried it). Is `tuple([] for i in range(dimension))` that bad?

matrix/test_unittest_matrix.py
rfire01: In general any class and function should have some documentation (the only exception is built in functions)

matrix/matrix.py
rfire01: its not that bad, its just I saw you did a lot of changes because of my comment, so I wanted to explain what I meant you to do (very small change)

matrix/matrix.py
```diff
@@ -41,12 +49,12 @@ def _validate(self, data: tuple):
     @classmethod
     def unity(cls, dimension: int):
         """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
-        unity_matrix = []
+        unity_matrix = ()
 
         for i in range(dimension):
-            unity_matrix.append((0,) * i + (1,) + (0,) * (dimension - i - 1))
+            unity_matrix += ((0,) * i + (1,) + (0,) * (dimension - i - 1),)
```
rfire01: why not with list comprehension?
(Last comment, fix it and the PR is closed)

yonatanevantal: You told me to seperate it. [https://github.com/osherdp/python-training/pull/60#discussion_r399812941](url)

matrix/matrix.py
```diff
@@ -49,12 +49,8 @@ def _validate(self, data: tuple):
...
     def unity(cls, dimension: int):
         """Generate a matrix with a main diagonal of 1's, elsewhere 0's."""
-        unity_matrix = ()
-
-        for i in range(dimension):
-            unity_matrix += ((0,) * i + (1,) + (0,) * (dimension - i - 1),)
-
-        return cls(unity_matrix)
+        return cls(tuple((0,) * i + (1,) + (0,) * (dimension - i - 1) for i
+                         in range(dimension)))
```
rfire01, line 53: I noticed you have a tendency to go from one extreme to another.
it is ok to do something in the middle to make it more readable
e.g.
lines = tuple(...)
return cls(lines)

PR #65, opened 2020-03-25
Stam exercise solved

galyammaya:


basic/stam/stamExcersize/stam.py
```diff
@@ -0,0 +1,49 @@
+__author__ = 'Maya Gal-Yam'
```
osherdp: Delete this and disable this in your IDE
Usually code is edited by lots of people

basic/stam/stamExcersize/stam.py
```diff
@@ -0,0 +1,49 @@
...
+    newfile_path = path.join(os.getcwd(), NEW_FILE_NAME)
+    file_to_create = open(newfile_path, 'w')
+    return file_to_create
+
+
+def manipulate(original, new):
+    for line in original.readlines():
+        new_line = NEW_LINE_FORMAT.format(line.strip())
+        new.writelines(new_line)
+    original.close()
```
osherdp: The convention is that the function who opens a file is the one to close it
It's to enable easy cleanup actions, and to prevent accessing closed files by mistake

basic/stam/stamExcersize/stam.py
```diff
@@ -0,0 +1,49 @@
...
+ORIGINAL_FILE_LOCATION = 1
+NEW_LINE_FORMAT = 'Maya {} Gal-Yam\n'
+NEW_FILE_NAME = 'newfile.txt'
+
+
+def get_file():
+    """gets the files passed in if args are valid"""
+    try:
+        originalfile_path = sys.argv[ORIGINAL_FILE_LOCATION]
+        if path.exists(originalfile_path) and path.isfile(originalfile_path):
```
osherdp: `path.exists` is redundant, `path.isfile` also checks if the file exists

Also, the whole checking is redundant, as the same thing will happen in`open` (but with fewer lines)

basic/stam/stamExcersize/stam.py
```diff
@@ -0,0 +1,49 @@
...
+
+ORIGINAL_FILE_LOCATION = 1
+NEW_LINE_FORMAT = 'Maya {} Gal-Yam\n'
+NEW_FILE_NAME = 'newfile.txt'
+
+
+def get_file():
+    """gets the files passed in if args are valid"""
+    try:
+        originalfile_path = sys.argv[ORIGINAL_FILE_LOCATION]
```
osherdp: Handle command line options (`sys.argv`) at one place, preferably in the top level (`if __name__ ==... `)

osherdp (changes requested):

basic/stam/stamExcersize/stam.py
```diff
@@ -0,0 +1,57 @@
...
+        new.writelines(new_line)
+
+
+def main():
+    try:
+        original_file = open(originalfile_path, 'r')
+        new_file = open(newfile_path, 'w')
+        manipulate(original_file, new_file)
+        original_file.close()
+        new_file.close()
```
osherdp: It's recommended to use the `with` keyword. For example:

```python
with open(originalfile_path, 'r') as original_file:
    with open(newfile_path, 'w') as new_file:
        manipulate(....
```

It makes sure that cleanup actions (like `close`) always happen
It's also called context-manager, and it's very recommended to use it

basic/stam/stamExcersize/stam.py
```diff
@@ -0,0 +1,57 @@
...
+    except FileNotFoundError:
+        print("Given path isn't a file")
+        sys.exit()
+
+
+if __name__ == '__main__':
+    try:
+        originalfile_path = sys.argv[ORIGINAL_FILE_LOCATION]
+        newfile_path = path.join(os.getcwd(), NEW_FILE_NAME)
```
osherdp: I think that the join with `os.getcwd()` is redundant, you can just use `NEW_FILE_NAME`

basic/stam/stamExcersize/stam.py
```diff
@@ -0,0 +1,57 @@
...
+    except IndexError:
+        print('No path given')
+        sys.exit()
```
osherdp:
```suggestion
        raise ValueError("No path given")
```

osherdp (changes requested):

basic/stam/stamExcersize/stam.py
galyammaya: I tried this but it failed the tests so I switched back

ranbentzvi: why the test fail?

galyammaya: Ok nevermind I checked and it wasn't related. I'll fix it

PR #66, opened 2020-03-25
adi's solution for stam exercise

AdiKadarLevi:
my solution

### DIFF
--- a/stam.py
+++ b/stam.py
@@ -6,13 +6,15 @@ LAST_NAME = 'Kadar Levi'
 
 def main():
     try:
-        input_file_path = input('please enter text file path\n')
+        input_file_path: str = input('please enter text file path\n')
         with open(input_file_path, 'r') as input_file:
             with open('new_text.txt', 'w+') as new_file:
                 for line in input_file:
-                    new_line = FIRST_NAME + ' ' + format(line[:-1]) + ' ' + LAST_NAME
+                    new_line = ' {} {} {}'.format(FIRST_NAME, line.strip(),  LAST_NAME)
                     print(new_line)
-                    new_file.write(new_line + '\n')
+                    new_file.writelines(new_line)
+    except FileNotFoundError:
+        print('Empty Path')
     except Exception as e:
         print(e)
 
--- a/stam.py
+++ b/stam.py
@@ -1,17 +1,20 @@
 __author__ = 'Adi'
 
+FIRST_NAME = 'Adi'
+LAST_NAME = 'Kadar Levi'
+
 
 def main():
-    input_file = open(r'C:\..., text.txt', 'r')
-    new_file = open('new_text.txt', 'w+')
-    # line = None
-    for line in input_file:
-        new_line = 'Adi ' + line[:-1] + ' Kadar Levi'
-        print(new_line)
-        new_file.write(new_line + '\n')
-
-    input_file.close()
-    new_file.close()
+    try:
+        input_file_path = input('please enter text file path\n')
+        with open(input_file_path, 'r') as input_file:
+            with open('new_text.txt', 'w+') as new_file:
+                for line in input_file:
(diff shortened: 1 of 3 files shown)


stam.py
```diff
@@ -0,0 +1,18 @@
+__author__ = 'Adi'
+
+
+def main():
+    input_file = open(r'C:\..., text.txt', 'r')
```
ranbentzvi: you should get the file path as an input from the user

ranbentzvi: read about `with` for working with files

stam.py
```diff
@@ -0,0 +1,18 @@
...
+    new_file = open('new_text.txt', 'w+')
+    # line = None
+    for line in input_file:
+        new_line = 'Adi ' + line[:-1] + ' Kadar Levi'
```
ranbentzvi: use consts instead of this magic strings

ranbentzvi: what would happen on an empty line?

ranbentzvi: also read about string `format`  function

ranbentzvi (changes requested):

stam.py
AdiKadarLevi: done

AdiKadarLevi: done

AdiKadarLevi: done

stam.py
AdiKadarLevi: done

AdiKadarLevi: done

stam.py
```diff
@@ -1,17 +1,20 @@
...
-    new_file.close()
+def main():
+    try:
+        input_file_path = input('please enter text file path\n')
+        with open(input_file_path, 'r') as input_file:
+            with open('new_text.txt', 'w+') as new_file:
+                for line in input_file:
+                    new_line = FIRST_NAME + ' ' + format(line[:-1]) + ' ' + LAST_NAME
+                    print(new_line)
+                    new_file.write(new_line + '\n')
```
ranbentzvi: you can use `writeline` and then you don't have to add the `\n` by yourself

stam.py
```diff
@@ -1,17 +1,20 @@
...
 
-    input_file.close()
-    new_file.close()
+def main():
+    try:
+        input_file_path = input('please enter text file path\n')
+        with open(input_file_path, 'r') as input_file:
+            with open('new_text.txt', 'w+') as new_file:
+                for line in input_file:
+                    new_line = FIRST_NAME + ' ' + format(line[:-1]) + ' ' + LAST_NAME
```
ranbentzvi: you should use the `format `function for the string creation instead of using `+`.
also, use `strip` for the `\n` removal

stam.py
```diff
@@ -1,17 +1,20 @@
...
+                    print(new_line)
+                    new_file.write(new_line + '\n')
+    except Exception as e:
```
ranbentzvi, line 18: try to catch an exact exception and not all of them

ranbentzvi, line 18: try to catch an exact exception and not all of them

ranbentzvi (changes requested):

stam.py
AdiKadarLevi: done

stam.py
AdiKadarLevi: done

stam.py
AdiKadarLevi: done

stam.py
```diff
@@ -10,11 +10,11 @@ def main():
         with open(input_file_path, 'r') as input_file:
             with open('new_text.txt', 'w+') as new_file:
                 for line in input_file:
-                    new_line = ' {} {} {}'.format(FIRST_NAME, line.strip(),  LAST_NAME)
+                    new_line = f" {FIRST_NAME} {line.strip()} {LAST_NAME}"
                     print(new_line)
                     new_file.writelines(new_line)
```
ranbentzvi, line 15: should be `writeline` , other than that perfect :)

PR #71, opened 2020-03-26
Feature/stam

idans124:


Pr1/stam.py
```diff
@@ -0,0 +1,41 @@
...
+def receive_first_name():
+    return input('Please enter your first name: ')
+
+
+def receive_last_name():
+    return input('Please enter your last name: ')
+
+
+def main():
+    file_name = sys.argv[FILE_NAME]
```
ranbentzvi, line 34: please use a meaningful name like `FILE_NAME_ARG_NUM`, although in this case, a magic number of 1 will be fine

Pr1/stam.py
```diff
@@ -0,0 +1,41 @@
...
+    first_name = receive_first_name()
+    last_name = receive_last_name()
+    write_to_file(file_name, FILE_NAME_NEW, first_name, last_name)
```
ranbentzvi, line 37: the new file name should be related to the original file

Pr1/stam.py
```diff
@@ -0,0 +1,41 @@
...
+                # print(line.strip())
+                line = line.strip()
+                with open(FILE_NAME_NEW, 'a') as new_file:
+                    new_file.write(first_name + ' ' + line.replace("\n", " ")
+                                   + ' ' + last_name + '\n')
+    except Exception as err:  # if stam.txt doesn't exist
+        print(err)  # print the error
+
+
+def receive_first_name():
```
ranbentzvi, line 25: functions are nice, but in this case, they contain only one function and used only once.
you can remove them and move their code to the main function.

Pr1/stam.py
```diff
@@ -0,0 +1,41 @@
+import sys
+
+
+FILE_NAME = 1
+FILE_NAME_NEW = "D:/Python Training/Pr1/stam_new.txt"
+
+
+def write_to_file(file_name, FILE_NAME_NEW, first_name, last_name):
```
ranbentzvi, line 8: You should not use a const as an argument to your function.
 if it's a const why have it as an argument?

Pr1/stam.py
```diff
@@ -0,0 +1,41 @@
...
+    """ Read from file stam.txt line by line
+     and write to file stam_new.txt with the first and last name
+    """
+
+    try:  # if stam.txt exist
+        with open(file_name, 'r') as orig_data:
+            for line in orig_data:
+                # print(line.strip())
```
ranbentzvi, line 16: remove unnecessary comment out code

Pr1/stam.py
```diff
@@ -0,0 +1,41 @@
...
+                line = line.strip()
+                with open(FILE_NAME_NEW, 'a') as new_file:
+                    new_file.write(first_name + ' ' + line.replace("\n", " ")
```
ranbentzvi, line 19: read about `format` function of string

Pr1/stam.py
```diff
@@ -0,0 +1,41 @@
...
+                                   + ' ' + last_name + '\n')
```
ranbentzvi, line 20: you can use `writeline` function and then you don't have to add the `\n` yourself

ranbentzvi (changes requested):

PR #75, opened 2020-03-27
Feature/atm project

idans124:


Pr1/stam.py
```diff
@@ -0,0 +1,41 @@
+import sys
```
osherdp, line 1: this is the ATM exercise, so you should remove this non-relevant file from the branch

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+        return dictionary
+    except Exception as error:
+        print(error)
+
+
+def main():
+    """
+    1) call read_file(file_name) to create dictionary
+    2) call who_are_you(dictionary) to recognize the user
+    3) call enter_password(dictionary, consumer_id)
```
osherdp: documentation is for human beings to use, so it's usually less technical ("call read_file(...)") but rather comprehensive ("loading DATABASE from filesystem")

Also, it doesn't make sense that `read_file` creates a directory :)

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+BALANCE = 2
+
+
+def update_file(file_name, dictionary):
+    """ update the file atm.txt with all the changes in dictionary
+    """
+    with open(file_name, 'w') as atm_file:
+        for key in dictionary:
+            temp_dict = dictionary[key]  # temp_dict = {password: balance}
+            for pw, bal in temp_dict.items():
```
osherdp: You can do:
```python
for key, (pw, bal) in dictionary.items():
    ...
```

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+                atm_file.write('{} {} {}\n'.format(key, pw, bal))
```
osherdp: I encourage you to use `password` and `balance` instead of `pw` and `bal`, for your code to be more understandable and more English-like

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+PASSWORD = 1
+BALANCE = 2
+
+
+def update_file(file_name, dictionary):
+    """ update the file atm.txt with all the changes in dictionary
+    """
+    with open(file_name, 'w') as atm_file:
+        for key in dictionary:
+            temp_dict = dictionary[key]  # temp_dict = {password: balance}
```
osherdp: `{password: balance}` is a weird way to represent the data
why not `[password, balance]`?

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+    Ask for a new password and change the dictionary accordingly
+    :param dictionary: old_dictionary
+    :param consumer_id: the current user's ID
+    :return: the updated dictionary
+    """
+    new_pass = input('Please enter a new password: ')
+    temp_dict = dictionary[consumer_id]  # temp_dict = {password: balance}
+    for password, balance in temp_dict.items():
+        temp_dict[new_pass] = temp_dict.pop(password)  # update password
```
osherdp: usually changing the dictionary (in this case `temp_dict` while iterating on is not recommended and might lead to undefined behavior

In this case, there's only one (key, value) pair so it won't be a problem. Anyway you should change the dictionary structure as it will make the code much simpler (talk to me via Whatsapp if you have more questions)

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+    dictionary[consumer_id] = temp_dict  # update the dictionary
```
osherdp: Comments should be used only in cases where it's very unclear what the code does
Roughly I would say one out of 100 lines of code should have a comment, not more

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+    Ask for the amount to deposit and change the dictionary accordingly
+    If the user enters a negative amount, the ATM informs it
+    :param dictionary: old_dictionary
+    :param consumer_id: the current user's ID
+    :return: the updated dictionary
+    """
+    cash = input('Please enter the amount of money you wish to deposit: ')
+    cash = int(cash)
+    if cash < 0:
+        print('Can''t do this, please enter a positive number!')
```
osherdp: I suppose it doesn't print it very good
you can do:
```python
        print("Can't do this, please enter a positive number!")
```

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+    Ask for the amount to withdraw and change the dictionary accordingly
+    If the user enters a negative amount, the ATM informs it
+    :param dictionary: old_dictionary
+    :param consumer_id: current user's ID
+    :return: the updated dictionary
+    """
+    cash = input('Please enter the amount of money you wish to withdraw: ')
+    cash = int(cash)
+    if cash < 0:
+        print('Can''t do this, please enter a positive number! ')
```
osherdp: same as above

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+    Go through the dictionary and find the user's balance
+    :param dictionary: dictionary
+    :param consumer_id: current user's ID
+    :return: the balance of the account
+    """
+    temp_dict = dictionary[consumer_id]  # temp_dict = {password: balance}
+    for password, balance in temp_dict.items():
+        return balance
+    return None  # the program never reaches to this point
```
osherdp: after changing the data structure you can delete this :)

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+                               '3-- cash deposit\n' +
+                               '4 -- change password\n' +
+                               'STOP --  if you are done\n')
+                if action == '1':
+                    balance = check_the_balance(dictionary, consumer_id)
+                    print('Your balance is: {}'.format(balance))
+
+                elif action == '2':
+                    dictionary = cash_withdrawal(dictionary, consumer_id)
+                    update_file(file_name, dictionary)
```
osherdp, line 123: this pattern also called switch-case. It's where based on some conditions you apply some actions
In Python, you can do the following:
```python
OPTIONS = {
    "1": check_the_balance,
    "2": cash_withdrawal,
    "3": ...
}
```
then in this function you can do the following:
```python
action = input("Press:...")
action_function = ACTIONS[action]
action_function(dictionary, consumer_id)
```

This will make the relevant function to be called, as you do the same things in all functions (running the function, updating the DB, etc.)
You might need to change the `check_the_balance` function to just print the balance by itself

ATM/atm.py
```diff
@@ -0,0 +1,202 @@
...
+    :return:
+    """
+    file_name = sys.argv[FILE_NAME]
+    dictionary = read_file(file_name)
+    while True:
+        consumer_id = who_are_you(dictionary)
+        if consumer_id == '-1':
+            print('Turning off the ATM...')
+            break
+        if enter_password(dictionary, consumer_id) is True:
```
osherdp: You might be better with the following:
```python
if not enter_password(...):
    continue

# Do the rest, knowing that the user has been successfully logged in

osherdp: `enter_password` is a bit misleading
maybe `verify_password` or `check_password`?

osherdp (changes requested):
Not bad at all
I think some changes are recommended. Talk with your tutor or me if you encounter any problems with those changes

PR #77, opened 2020-03-29
stam

ghost:


stam.py
```diff
@@ -0,0 +1,20 @@
...
+    file_to = r'C:\Users\nimro\Documents\python_test\homework_name.txt'
+
+    while True:
+        file_name = input('enter file-  ')
+
+        if os.path.isfile(file_name):
+            read_file = open(file_name, 'r')
+            for line in read_file:
+                with open(file_to, 'a') as write_file:
+                    write_file.write('nimrod    ' + line.replace('\n', '') + '    satt' + '\n')
```
snudler6, line 14: You can use f-strings instead of string concatenation
See here: http://zetcode.com/python/fstring/

stam.py
```diff
@@ -0,0 +1,20 @@
...
+def main():
+    file_to = r'C:\Users\nimro\Documents\python_test\homework_name.txt'
+
+    while True:
+        file_name = input('enter file-  ')
+
+        if os.path.isfile(file_name):
+            read_file = open(file_name, 'r')
+            for line in read_file:
+                with open(file_to, 'a') as write_file:
```
snudler6, line 13: If you would prefer the output file to be cleaned before writing to it, you can change the open format to `'w'`

stam.py
```diff
@@ -0,0 +1,20 @@
+import os
+
```
snudler6, line 2: Your code doesn't pass a Coding convention called flake8.
In coding, using a well established coding conventions will help others to read your code.
check you comments here: https://github.com/osherdp/python-training/pull/77/checks

snudler6 (changes requested):

snudler6:
All in all, nice clean and minimal work 👍

PR #93, opened 2020-03-30
solved matrix exercise 1

danieljprince:


matrix.py
```diff
@@ -0,0 +1,255 @@
+"""
+Matrix exercise, Daniel Prince - unity training
+"""
```
rfire01: should be 1 line (fix everywhere)

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+class Matrix:
+
```
rfire01, line 6: doc?

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+    def __init__(self, data):
+        """
+        initializes the entered matrices and makes sure it is n x n
+        """
+        self.data = data
+        # Check if n x n
+        for tup in self.data:
```
rfire01: give meaningful names

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+            if len(self.data) != len(tup):
+                raise ValueError("not a valid matrix. (not nXn)")
+            else:
+                continue
```
rfire01: fix file according to pep 8

rfire01: why?

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def __str__(self):
+        """display just the actual value of the instance
+
+        Returns: the main outside tuple
+        """
```
rfire01: No need for doc in built in functions

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+        return str(self.data)
+
+    def __repr__(self):
+        """display the instance of the class as is.
+
+        Returns: a readable form of the matrix
+        """
+        return "Matrix({})".format(self.data)
```
rfire01: What will happen if you decide to change the class name?

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+        return iter(self.data)
+
+    def __add__(self, other):
+        """ Adds two matrices together.
+
+        Args: the first matrix (self)
+            other: the second matrix
+
+        Returns: a new Class-type final addition result matrix.
+
```
rfire01: remove redundant space at the start of the doc.
Remove blank line at the end.
Fix the entire doc of your code according to conventions
chpater 3.8 - https://github.com/google/styleguide/blob/gh-pages/pyguide.md#38-comments-and-docstrings

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+        """
+        if len(self.data) == len(other.data):
```
rfire01: What will happen if I do
m=Matrix(...)
m + 1

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+            post_add_list = []
+            for tupA, tupB in zip(self.data, other.data):
+
```
rfire01, line 31: remove

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+                for numA, numB in zip(tupA, tupB):
+                    post_add_list.append(numA + numB)
```
rfire01: why not list comprehension?

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+
+            return Matrix(tuple(splitting(post_add_list, len(self.data))))
```
rfire01: Try to think about more pythonic way to implement this function

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+
+        else:
+            print("cannot perform addition on unequal matrix length")
+
+    def __sub__(self, other):
```
rfire01, line 55: Same as add

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+        """ Multiply two matrices or a matrix by a scalar
+
+        Args: original matrix (self)
+            other: a number or a second matrix
+
+        Returns: a new Class-type final multiplication result matrix.
+
+        """
+        new_list = []
+        if type(other) == Matrix:
```
rfire01: isinstance

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+
+                for numA, numB in zip(tupA, tupB):
+                    post_sub_list.append(numA - numB)
+
+            return Matrix(tuple(splitting(post_sub_list, len(self.data))))
+
+        else:
+            print("cannot perform subtraction on unequal matrix length")
+
+    def __mul__(self, other):
```
rfire01, line 64: find more pythonic way to do this

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+                # Inverting the second matrix
+                inverted_matrix = [] * len(self.data)
+                for i in range(len(self.data)):
+
+                    for tup in other.data:
+                        inverted_matrix.append(tup[i])
+
+                for t in self.data:
+                    for l in splitting(inverted_matrix, len(self.data)):
+                        for num, i2 in zip(l, range(len(self.data))):
```
rfire01: This is not readable. what is t? what is l? you made it complicated using variables with no meaning

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+                print("cannot perform multiplication on uneven matrices")
+
+        else:
+            for tup in self.data:
+                for num in tup:
+                    new_list.append(num * other)
+
+        return Matrix(tuple(splitting(new_list, len(self.data))))
+
+    def __floordiv__(self, other):
```
rfire01: make it pythonic

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+            for tup in self.data:
+                for num in tup:
+                    new_list.append(num / other)
+
+            return Matrix(tuple(splitting(new_list, len(self.data))))
+
+        else:
+            print("cannot divide matrix by matrix")
+
+    def __eq__(self, other):
```
rfire01: make it pythonic

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+        if type(other) == Matrix:
+            if self.data == other.data:
+                return True
+            else:
+                return False
+
+        else:
+            print("cannot check equivalency of type matrix and type int. ")
+
+    def __ne__(self, other):
```
rfire01: make it pythonic

matrix.py
```diff
@@ -0,0 +1,255 @@
...
+    def __hash__(self):
+        """ Make dict keys a hash
+
+        Returns: the hashed tuple
+
+        """
+        return hash(self.data)
+
+    @staticmethod
+    def unity(num):
```
rfire01: make it pythonic - do it for all functions and submit again.

matrix.py
```diff
@@ -0,0 +1,255 @@
+"""
```
rfire01: Where are the unitests?

rfire01 (changes requested):

matrix.py
natsala13: use fstring

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                                 + other.data[row_i][column_i]
+                                 for row_i in range(len(self.data))
+                                 for column_i in range(len(other.data))]
+                return Matrix(tuple(splitting(post_add_list, len(self.data))))
+
+            else:
+                print("matrices must be equal length")
+
+        else:
+            print("cannot add a scalar")
```
IgalKolihman: raise an exception instead of printing in the terminal

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                                 - other.data[row_i][column_i]
+                                 for row_i in range(len(self.data))
+                                 for column_i in range(len(other.data))]
+                return Matrix(tuple(splitting(post_sub_list, len(self.data))))
+
+            else:
+                print("matrices must be equal length")
+
+        else:
+            print("cannot subtract from a / a scalar")
```
IgalKolihman: same as my comment to `__add__`

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                for tup in self.data:
+                    for split_list in split_column_matrix:
+                        for num, index in \
+                                zip(split_list, range(len(self.data))):
+                            post_mul.append(num * tup[index])
+                split_mul = splitting(post_mul,
+                                      len(self.data) * len(self.data))
+                added = [sum(tuples) for tuples in split_mul]
+                final_mtx = splitting(added, len(self.data))
+                return Matrix(tuple(final_mtx))
```
IgalKolihman: when you have such complex code, add some comments that describe what is happening.

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+    @staticmethod
+    def unity(num):
+        """Display tuple of tuples with 1 & 0
+        so that the ones create a diagonal if seen as a square.
+
+            Args:
+                num: the amount of ones the user wants
+
+            Returns: the final tuple, as a Matrix type
```
IgalKolihman: fix indentation

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        if not isinstance(other, Matrix):
+            divided_mtx = [num / other for tup in self.data for num in tup]
+            return Matrix(tuple(splitting(divided_mtx, len(self.data))))
+
+        else:
+            print("cannot divide matrix by matrix")
+
+    @staticmethod
+    def unity(num):
+        """Display tuple of tuples with 1 & 0
```
IgalKolihman:
```suggestion
        """Display tuple of tuples with 1 & 0.
```

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        so that the ones create a diagonal if seen as a square.
+
+            Args:
+                num: the amount of ones the user wants
```
IgalKolihman: specify the type of num

matrix.py
IgalKolihman: in the `Returns` we first write the type of the returned value, and then the description of the value

```suggestion
            Returns: tuple. The final tuple, as a Matrix type
```

and it should be bellow the `Returns` title:
```suggestion
            Returns:
                tuple. The final tuple, as a Matrix type
```

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+            a_list: the original unsliced list
+            slice_number: the amount of small lists post slicing
+
+        Returns: the final list of smaller tuples
+    """
+
+    inverted_list = [tuple(a_list[i * len(a_list) // slice_number:
+                                  (i + 1) * len(a_list) // slice_number])
+                     for i in range(slice_number)]
+    return inverted_list
```
IgalKolihman, line 153: check out this option for splitting your list into smaller lists:
https://stackoverflow.com/a/312464/5898898

it looks cleaner and simpler

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        """Display tuple of tuples with 1 & 0
+        so that the ones create a diagonal if seen as a square.
+
+            Args:
+                num: the amount of ones the user wants
+
+            Returns: the final tuple, as a Matrix type
+        """
+
+        unity_list = [0 for _ in range(num * num)]
```
IgalKolihman: why not
```suggestion
        unity_list = [0] * (num*num)
```

**do the same to line 128**

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+    @property
+    def tuples(self):
+        """Display the instance as tuple.
+
+            Returns: the raw tuple
+        """
+
+        return self.data
+
+    # END OF CLASS --------------
```
IgalKolihman: remove

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+    @staticmethod
+    def ones(num):
+        """Display a tuple of tuples filled with ones.
+
+            Args:
+                num: the amount of tuples the user wants
+
+            Returns: the final tuple, as a Matrix type
+        """
```
IgalKolihman: fix your indentation in your docs around the code

 ```suggestion
        """Display a tuple of tuples filled with ones.

        Args:
            num: the amount of tuples the user wants

        Returns: the final tuple, as a Matrix type
        """
```

File: matrix.py [hunk 31]
```diff
@@ -0,0 +1,156 @@
...
+
+        ones_list = [1 for _ in range(num * num)]
+        return Matrix(tuple(splitting(ones_list, num)))
+
+    @property
+    def tuples(self):
+        """Display the instance as tuple.
```
IgalKolihman: more like
```suggestion
        """Return the instance as tuple.
```

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # END OF CLASS --------------
+
+
+def splitting(a_list, slice_number):
+    """Split a list into smaller lists and convert them to tuples.
+
+        Args:
+            a_list: the original unsliced list
+            slice_number: the amount of small lists post slicing
```
IgalKolihman: specify types for all arguments that you get in all your functions in all docs

```suggestion
            a_list (list): the original unsliced list
            slice_number (number): the amount of small lists post slicing
```

test_matrix.py
```diff
@@ -0,0 +1,78 @@
+import unittest
```
IgalKolihman, line 3: add page documentation

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+from matrix import Matrix
+
+
+class TestMatrix(unittest.TestCase):
```
IgalKolihman, line 7: add some negative tests which check that your class raises errors when needed.
_i.e: an exception should raise if the items in the tuple are `str` or `bool`_

IgalKolihman (changes requested):

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        # Check if n x n
+        for tup in self.data:
+            if len(self.data) != len(tup):
+                raise ValueError("not a valid matrix. (not nXn)")
+
+    def __str__(self):
+        return str(self.data)
+
+    def __repr__(self):
+        return '{self.__class__.__name__}({self.data})'.format(self=self)
```
natsala13: use fstring instead of string formatting.

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        return iter(self.data)
+
+    def __hash__(self):
+        return hash(self.data)
+
+    def __eq__(self, other):
+        return self.data == other.data
+
+    def __ne__(self, other):
+        return self.data != other.data
```
natsala13: Try using already implemented methods instead of implementing it.

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __add__(self, other):
+        """Adds 2 matrices."""
+
+        if isinstance(other, Matrix):
+            if len(self.data) == len(other.data):
```
natsala13: If you are using len(self.data) that much, consider saving it as a class property.

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                post_add_list = [self.data[row_i][column_i]
+                                 + other.data[row_i][column_i]
+                                 for row_i in range(len(self.data))
+                                 for column_i in range(len(other.data))]
```
natsala13: Instead of iterating through the array's length and accessing every element every time, iterate through the array itself. 
For example if you wish to print an array you could do:
`for i in range(len(array)): print(array[i])`

Or you could simply do:
`for val in array: print(val)`

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                return Matrix(tuple(splitting(post_add_list, len(self.data))))
```
natsala13: As a general comment,
Instead of creating a list and converting it to tuple, create it as a tuple from the beginning.
remove splitting method its unnecessary.
Learn how to create 2 dim tuples with list comprehension.

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+            else:
+                print("matrices must be equal length")
+
+        else:
+            print("cannot subtract from a / a scalar")
+
+    def __mul__(self, other):
+        """Multiplies matrix by scalar/matrix."""
+
+        if isinstance(other, Matrix):
```
natsala13, line 65: Try dividing it to smaller methods, it will make you code clearer.

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                final_mtx = splitting(added, len(self.data))
+                return Matrix(tuple(final_mtx))
+            else:
+                print("cannot multiply different dimension matrices")
+
+        else:
+            final_mtx = [num * other for tup in self.data for num in tup]
+            return Matrix(tuple(splitting(final_mtx, len(self.data))))
+
+    def __truediv__(self, other):
```
natsala13: Try using already implemented methods.

natsala13 (changes requested):

matrix.py
danieljprince: That is good advice,  although now that its done, should I really go back and do it? its seems that it wouldn't be as officiant.
I would have to call the class property every time and it doesn't seem to save much time and effort considering It works. 
But like I said, I wish I would have thought of that before.
I hope to implement it to my next code.

matrix.py
danieljprince: I thought of that originally, but the list I need to iterate over is made up of smaller lists.
so looping over the elements themselves wouldn't give me the number I need, it would give me a list. 
this can still be done but I was told to use list comprehension and I don't know of a way of doing it using list comprehension.

matrix.py
danieljprince: will do!
although in my mul function, the best way I found was by splitting a list, so I decided 
to use it to convert to tuple as well. instead of writing the same code a bunch of times.

matrix.py
danieljprince: Even though its not the end of the sentence? or are you proposing I should get rid of the rest of the sentence?

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        else:
+            print("cannot subtract from a / a scalar")
+
+    def __mul__(self, other):
+        """Multiplies matrix by scalar/matrix."""
+
+        if isinstance(other, Matrix):
+            if len(self.data) == len(other.data):
+                # Converting the second matrix to columns
+                column_matrix = [tup[index] for index in range(len(self.data))
```
jhb1996, line 72: If you are grabbing every entry in tup in order then you don't need to use range and can just loop directly over the entries of tup

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                post_mul = []
+                for tup in self.data:
+                    for split_list in split_column_matrix:
+                        for num, index in \
+                                zip(split_list, range(len(self.data))):
+                            post_mul.append(num * tup[index])
+                split_mul = splitting(post_mul,
+                                      len(self.data) * len(self.data))
+                added = [sum(tuples) for tuples in split_mul]
+                final_mtx = splitting(added, len(self.data))
```
jhb1996: We dont use abbreviations in names (here and everywhere).

matrix.py
jhb1996: They way to write docs is that your first line should always be a complete sentence giving a super brief description. The next block (if it is needed) should be the beginning of a new sentence describing the function in more detail.

So rewrite the first line to summarize the function and if you need a more detailed description then give it on the next line in one or more new full sentances.

matrix.py
jhb1996: Pretty sure our convention is to write "number" for all numbers

matrix.py
natsala13: The purpose of this exercise is to train writing python. I think that if you spend some time on rewriting code even if it doesn't make it more efficient, you don't lose anything.
Anyway I am quite sure its a 5 minutes change lol.

matrix.py
natsala13: You are right.
Use the function `zip`. it let you iterate over two arrays in parallel. (google it if its not familiar)
Using that you would be able to write a two-dim for loop and accessing it element by element without indexing.

matrix.py
natsala13: Don't worry, we'll find a way to skip it there to ;)

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        """Display a tuple of tuples filled with ones.
+
+            Args:
+                num: the amount of tuples the user wants
+
+            Returns: the final tuple, as a Matrix type
+        """
+
+        ones_list = [1 for _ in range(num * num)]
+        return Matrix(tuple(splitting(ones_list, num)))
```
natsala13: Create it using a two dim list comprehension.

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                print("matrices must be equal length")
+
+        else:
+            print("cannot subtract from a / a scalar")
+
+    def __mul__(self, other):
+        """Multiplies matrix by scalar/matrix."""
+
+        if isinstance(other, Matrix):
+            if len(self.data) == len(other.data):
```
natsala13: Lets find step by step a better way writing this code using list comprehension.

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                # Converting the second matrix to columns
+                column_matrix = [tup[index] for index in range(len(self.data))
+                                 for tup in other.data]
+                split_column_matrix = splitting(column_matrix, len(self.data))
+                post_mul = []
+                for tup in self.data:
+                    for split_list in split_column_matrix:
+                        for num, index in \
+                                zip(split_list, range(len(self.data))):
+                            post_mul.append(num * tup[index])
```
natsala13: I believe that what you intended to do here, is given two vectors lets multiply it element by element.
First you could iterate over `zip(split_list, tup)` instead over iterating over a range and accessing it using indexing.
Second you could replace this part by a one line list comprehension.
Finally you might as well sum it, it will save you work later splitting it s=to smaller lists in order to sum these exact values.

I believe after that you would be able to change the rest of the method so it would look more pythonic.

natsala13 (changes requested):

matrix.py
```diff
@@ -0,0 +1,156 @@
+"""Matrix exercise, Daniel Prince - unity training."""
```
rfire01: Still should be file doc, no need to add your name or that its training :)

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+
```
rfire01, line 3: remove

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+class Matrix:
+    """Perform all sorts of matrix operations."""
+
```
rfire01, line 6: remove

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self, data):
+        self.data = data
+        # Check if n x n
+        for tup in self.data:
+            if len(self.data) != len(tup):
+                raise ValueError("not a valid matrix. (not nXn)")
```
rfire01: Is it the only case that the input is not a valid matrix?

matrix.py
```diff
@@ -0,0 +1,156 @@
+"""Matrix exercise, Daniel Prince - unity training."""
+
+
+class Matrix:
+    """Perform all sorts of matrix operations."""
+
+    def __init__(self, data):
+        self.data = data
+        # Check if n x n
+        for tup in self.data:
```
rfire01: Use meaningful names (call a variable tup, list, dict etc. says what type it is but not what it represents)

matrix.py
rfire01: use fstring

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        return hash(self.data)
+
+    def __eq__(self, other):
+        return self.data == other.data
+
+    def __ne__(self, other):
+        return self.data != other.data
+
+    def __add__(self, other):
+        """Adds 2 matrices."""
```
rfire01: You don't have to write doc for built-in functions, but if you already did write the doc according to conventions

matrix.py
rfire01: Actually I don't see the advantage of using len property instead of len() as you calculate it every time again. 
If you want "save" something, then use cached_property, through for length it doesn't change much

matrix.py
rfire01: Usually you should try to call the class explicitly.
What will happen if you change the class name? you will need to change it everywhere you use it.
You can call it implicitly and this way solve the issue.
Fix it everywhere you did it

matrix.py
rfire01: Who told you other is scalar?

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        if isinstance(other, Matrix):
+            if len(self.data) == len(other.data):
+                post_add_list = [self.data[row_i][column_i]
+                                 + other.data[row_i][column_i]
+                                 for row_i in range(len(self.data))
+                                 for column_i in range(len(other.data))]
+                return Matrix(tuple(splitting(post_add_list, len(self.data))))
+
+            else:
+                print("matrices must be equal length")
```
rfire01: you can give a little more details, such as what length of each matrix :)

matrix.py
rfire01: Instead of all of this try to use functions you already created in order to implement this functions

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+        else:
+            print("cannot subtract from a / a scalar")
+
+    def __mul__(self, other):
+        """Multiplies matrix by scalar/matrix."""
+
```
rfire01: remove

matrix.py
rfire01: if you do matrix validation in several functions you can use decorator to avoid code duplication

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                split_mul = splitting(post_mul,
+                                      len(self.data) * len(self.data))
+                added = [sum(tuples) for tuples in split_mul]
+                final_mtx = splitting(added, len(self.data))
+                return Matrix(tuple(final_mtx))
+            else:
+                print("cannot multiply different dimension matrices")
+
+        else:
+            final_mtx = [num * other for tup in self.data for num in tup]
```
rfire01: why not create from the start tuple inside of a tuple?

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __truediv__(self, other):
+        """Divide matrix by scalar."""
+
+        if not isinstance(other, Matrix):
+            divided_mtx = [num / other for tup in self.data for num in tup]
+            return Matrix(tuple(splitting(divided_mtx, len(self.data))))
+
+        else:
+            print("cannot divide matrix by matrix")
```
rfire01: use already defined functions to implement

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+            else:
+                print("cannot multiply different dimension matrices")
+
+        else:
+            final_mtx = [num * other for tup in self.data for num in tup]
+            return Matrix(tuple(splitting(final_mtx, len(self.data))))
+
+    def __truediv__(self, other):
+        """Divide matrix by scalar."""
+
```
rfire01, line 92: remove in your code redundant blank lines

matrix.py
rfire01: 1. this function create matrix not tuples
2. this function returns a matrix, it doesn't display anything

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+            divided_mtx = [num / other for tup in self.data for num in tup]
+            return Matrix(tuple(splitting(divided_mtx, len(self.data))))
+
+        else:
+            print("cannot divide matrix by matrix")
+
+    @staticmethod
+    def unity(num):
+        """Display tuple of tuples with 1 & 0
+        so that the ones create a diagonal if seen as a square.
```
rfire01: First line of doc should describe what the function do (single line)

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+        if not isinstance(other, Matrix):
+            divided_mtx = [num / other for tup in self.data for num in tup]
+            return Matrix(tuple(splitting(divided_mtx, len(self.data))))
+
+        else:
+            print("cannot divide matrix by matrix")
+
+    @staticmethod
+    def unity(num):
```
rfire01: same as before, avoid using type for names.
What num means? what it represents?

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+
+            Args:
+                num: the amount of ones the user wants
+
+            Returns: the final tuple, as a Matrix type
+        """
+
+        unity_list = [0 for _ in range(num * num)]
+        unity_list[::num + 1] = [1 for _ in unity_list[::num + 1]]
+        return Matrix(tuple(splitting(unity_list, num)))
```
rfire01: Find another way to do it, the usage of your splitting function is forbidden for it

matrix.py
rfire01: Notice that if there is a comment about something, that you fix this issue everywhere (I won't write the comment again everywhere)

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+        """
+
+        ones_list = [1 for _ in range(num * num)]
+        return Matrix(tuple(splitting(ones_list, num)))
+
+    @property
+    def tuples(self):
+        """Display the instance as tuple.
+
+            Returns: the raw tuple
```
rfire01: Fix according doc convetions

test_matrix.py
rfire01, line 3: page doc?

test_matrix.py
```diff
@@ -0,0 +1,78 @@
+import unittest
+from matrix import Matrix
```
rfire01, line 4: There should be blank line between them.
Notice conventions about imports:
---built in imports---
__blank line__
--third part imports--
__blank line__
--local methods--
__blank line__

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+    """Tests my matrix methods."""
+
+    def setUp(self):
+        self.a = Matrix(((1, 2), (3, 4)))
+        self.b = Matrix(((5, 6), (7, 8)))
+        self.c = Matrix(((1, 2, 3), (4, 5, 6), (7, 8, 9)))
+        self.d = Matrix(((2, 3, 4), (5, 6, 7), (7, 8, 9)))
+
+    def tearDown(self):
+        pass
```
rfire01: if you don't use it, don't define it

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_init(self):
```
rfire01, line 16: add doc to all tests

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+        with self.assertRaises(ValueError):
+            self.e = Matrix(((1, 2), (1, 2, 3)))  # not nXn
```
rfire01, line 18: This doesn't check init, it check init failure

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_add(self):
+        self.assertEqual(Matrix.__add__(self.a, self.b),
```
rfire01, line 21: This is not how you use built-in functions.
Fix the entire test file

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+                         Matrix(((6, 8), (10, 12))))
+        self.assertEqual(Matrix.__add__(self.c, self.d),
+                         Matrix(((3, 5, 7), (9, 11, 13), (14, 16, 18))))
+        # test addition of matrices with different dimensions
+        self.assertEqual(Matrix.__add__(self.a, self.c), None)
```
rfire01: You compare between two matrix objects, which means the _ _ eq _ _ function is used to do it,
This mean instead of checking a single functionality here you check 2.
This mean that you can have a bug in add, but the test won't fail because you also have bug in eq function (for example it always return true)

matrix.py
rfire01: Raise an error.
If you just print to the screen and don't raise an error it means that function finished correctly

test_matrix.py
rfire01: what about adding matrix with non matrix object?

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+    def test_init(self):
+        with self.assertRaises(ValueError):
+            self.e = Matrix(((1, 2), (1, 2, 3)))  # not nXn
+
+    def test_add(self):
+        self.assertEqual(Matrix.__add__(self.a, self.b),
+                         Matrix(((6, 8), (10, 12))))
+        self.assertEqual(Matrix.__add__(self.c, self.d),
+                         Matrix(((3, 5, 7), (9, 11, 13), (14, 16, 18))))
+        # test addition of matrices with different dimensions
```
rfire01: if you need to write a doc for it, it means that it should get its own test

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+        self.assertEqual(Matrix.__add__(self.a, self.c), None)
+
+    def test_sub(self):
+        self.assertEqual(Matrix.__sub__(self.a, self.b),
+                         Matrix(((-4, -4), (-4, -4))))
+        self.assertEqual(Matrix.__sub__(self.d, self.c),
+                         Matrix(((1, 1, 1), (1, 1, 1), (0, 0, 0))))
+        # test subtraction of matrices with different dimensions
```
rfire01: if you need to write a doc for it, it means that it should get its own test

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+        self.assertEqual(Matrix.__sub__(self.a, self.c), None)
+
+    def test_mul(self):
+        self.assertEqual(Matrix.__mul__(self.a, self.b),
+                         Matrix(((19, 22), (43, 50))))
+        self.assertEqual(
+            Matrix.__mul__(self.c, self.d),
+            Matrix(((33, 39, 45), (75, 90, 105), (117, 141, 165)))
+        )
+        # test multiplication of matrices with different dimensions
```
rfire01: if you need to write a doc for it, it means that it should get its own test, and everywhere else

test_matrix.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_sub(self):
+        self.assertEqual(Matrix.__sub__(self.a, self.b),
+                         Matrix(((-4, -4), (-4, -4))))
+        self.assertEqual(Matrix.__sub__(self.d, self.c),
+                         Matrix(((1, 1, 1), (1, 1, 1), (0, 0, 0))))
+        # test subtraction of matrices with different dimensions
+        self.assertEqual(Matrix.__sub__(self.a, self.c), None)
+
+    def test_mul(self):
```
rfire01, line 40: what about b *a (the other order)?
What about multiply with scalar?

matrix.py
```diff
@@ -0,0 +1,156 @@
...
+                                 for column_i in range(len(other.data))]
+                return Matrix(tuple(splitting(post_sub_list, len(self.data))))
+
+            else:
+                print("matrices must be equal length")
+
+        else:
+            print("cannot subtract from a / a scalar")
+
+    def __mul__(self, other):
```
rfire01, line 64: what about _ _rmul_ _?

rfire01 (changes requested):

matrix.py
danieljprince: It tells me pep8 needs 2 lines before a class. 
Remove anyway?

matrix.py
danieljprince: Last time you told me to add a doc string at the start of the class. 
Remove anyway?

rfire01: The remove talks about the blank line after the doc

matrix.py
rfire01: pep8 tells you need 2 lines before class (if there is something (non comment) before it).
Doc line doesn't count as something before it

Calculator_exercise.py
```diff
@@ -0,0 +1,299 @@
...
+def main():
+    """Get users equation and sends it through the 'chain' of solving.
+
+    Returns: str - The final result.
+    """
+
+    user_eq = input("Enter equation (type 'quit' to exit)\n")
+    if user_eq != 'quit':
+        print(space_remove(user_eq))
+        main()
```
jhb1996, line 291: When running a program continually do it with a while loop rather than with recursion. It is easier to read. 

Also you every time you make a recursive call (which is every you get an input) you take up a little bit of memory since the program needs to know where to return to if when the recursive function call ends. If your calculator ran for a really long time you would fill up your entire computer's memory with recursive calls.

Calculator_exercise.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+
+    user_eq = user_eq.replace(' ', '')
+    return parenthesis(user_eq)
+
+
+def main():
+    """Get users equation and sends it through the 'chain' of solving.
+
+    Returns: str - The final result.
```
jhb1996, line 285: This function doesn't actually return anything. (Printing is not considered returning)

Calculator_exercise.py
```diff
@@ -0,0 +1,299 @@
+"""Calculator exercise"""
```
jhb1996, line 1: You pushed this to "solved matrix exercise 1." I assume this is meant to be in its own branch (I'm not sure how you were told to do things).

Calculator_exercise.py
```diff
@@ -0,0 +1,299 @@
...
+
```
jhb1996, line 2: A general comment on your approach. You are calling a function ("space remover", which calls a "parenthesis" which calls factorial). The problem with this is that it makes finding something specific in your code very difficult for anyone reading it (and will also be way harder to debug).

Your code will be much clearer if you have "handoffs" within some main evaluate
To be clear. You want something like

def evaluate(expression):
         expression = remove_spaces(expression)
         expression = handle_parenthesis(expression)
         expression = handle_factorials(expression)

this allows me to easily see the various steps in your code and then I can look at the function implementation if I want more details.

You can also do a more object oriented approach. Something like

def calculate(input):
            my_expression = Expression(input)
            my_expression.remove_spaces()
            my_expression.remove_parentheses()
            my_expression.handle_factorials()

For this exercise (I think) both are considered valid approaches. In general we prefer the object oriented approach as it is cleaner and easier to follow when problems get bigger.

Calculator_exercise.py
```diff
@@ -0,0 +1,299 @@
...
+            matches2 = re.findall(r'-?\d+', match)
+            int_matches = [int(x) for x in matches2]
+            peace = peace.replace(match, str(max(int_matches)))
+        return maximum(peace)
+
+    else:  # No more '$' -
+        return minimum(peace)
+
+
+def average(peace):
```
jhb1996, line 189: Average should not get precedence over max or min. Whichever comes first (is more to the left) in the equation should be performed first. As you wrote it you will have a problem with 

3 $ 1 @ 3

 If you take the average before the max you will get different (incorrect) results.

You have the same issue with the other equal priority operations.

Note: This will take some major rewriting to fix. Do yourself a favor and write out the logic of how you plan to do this before you start coding it. There is a simple and clean way to handle this and an ugly and messy way.

test_calcuator.py
```diff
@@ -0,0 +1,57 @@
+"""Calculator Tests."""
```
jhb1996, line 1: You are missing many important test scenarios. It is almost always important not just to test things individually but to make sure they work together in combination. In particular for the calculator a major thing you need to test is that if you have multiple operations they get performed in the right order.

test_calcuator.py
```diff
@@ -0,0 +1,57 @@
...
+import Calculator_exercise
+
+
+class TestCalc(unittest.TestCase):
+    """Tests solving different equations"""
+
+    def setUp(self):
+        pass
+
+    def test_add(self):
```
jhb1996, line 13: For all of your tests you need to make sure you test negatives and decimals as well.

test_calcuator.py
```diff
@@ -0,0 +1,57 @@
+"""Calculator Tests."""
+
+import unittest
+import Calculator_exercise
```
jhb1996, line 4: rather than import calculator exercise and call every function individually you should call the highest level function which does calculations and check that the ultimate answer is correct.

For the way you currently wrote the code it means you should call remove_spaces() each test and make sure you get the correct result for each operation. 

Given the necessary restructuring that you need to do to calculator file you should avoid making this change until you do the restructuring.

jhb1996 (changes requested):

PR #102, opened 2020-03-30
atm excersize solved

galyammaya:


basic/atm/atm.py
```diff
@@ -0,0 +1,111 @@
...
+    for key in customers:
+        file.writelines(
+            CUSTOMER_LINE_FORMAT.format(customers[key][NAME_INDEX],
+                                        customers[key][PASSWORD_INDEX],
+                                        customers[key][BALANCE_INDEX],
+                                        key))
+
+
+def main():
+    with open(customers_file_path, 'r+') as customers_file:
```
ranbentzvi: the file should be opened only when you need to read/write to it and not to the whole time

basic/atm/atm.py
```diff
@@ -0,0 +1,111 @@
...
+        customers = get_customers(customers_file)
+        flag = True
+        while flag:
+            cust_ID = input(WELCOME_MESSAGE)
+            if cust_ID in customers.keys():
+                print('Hello {}'.format(customers[cust_ID][NAME_INDEX]))
+                user_input = input(MENU_MESSAGE)
+                if user_input == '1':
```
ranbentzvi: should be consts

basic/atm/atm.py
```diff
@@ -0,0 +1,111 @@
+import sys
```
ranbentzvi: you should create an atm class an out all the functionality inside this class

ranbentzvi (changes requested):

basic/atm/atm.py
galyammaya: the message or the input options or both?

basic/atm/framework.py
```diff
@@ -0,0 +1,112 @@
...
+#                                             customers[key][BALANCE_INDEX],
+#                                             key))
+
+
+def main():
+    atm = ATM()
+    customers = atm.get_customers(customers_file_path)
+    flag = True
+    while flag:
+        cust_ID = input(WELCOME_MESSAGE)
```
ranbentzvi, line 24: make this part of the atm class, and then you don't have to pass it to all your functions.
also, I think you can combine both files it seems like most of this file is commented out.

ranbentzvi (changes requested):

basic/atm/framework.py
```diff
@@ -17,91 +17,30 @@
...
     while flag:
         cust_ID = input(WELCOME_MESSAGE)
-        if cust_ID in customers.keys():
-            print(HELLO_MESSAGE.format(customers[cust_ID][NAME_INDEX]))
+        if cust_ID in atm.customers.keys():
+            print(HELLO_MESSAGE.format(atm.customers[cust_ID][NAME_INDEX]))
             user_input = input(MENU_MESSAGE)
             if user_input == MENU_CHECK:
-                atm.check_balance(customers, cust_ID)
+                atm.check_balance(cust_ID)
```
ranbentzvi: if you would make the customer ID class variable everything would be nicer

ranbentzvi (changes requested):

ranbentzvi (approved):

PR #106, opened 2020-03-30
adi's first solution for ATM

AdiKadarLevi:


ATM.py
```diff
@@ -0,0 +1,120 @@
...
+        check_password(id_num, password, pass_dict)
+        print('What would you like to do today?\n'
+              'To check your balance press 1\n'
+              'To deposit money press 2\n'
+              'To withdraw money press 3\n'
+              'To change password press 4\n'
+              'for Exit press -1\n')
+        choose = input('\n')
+        while choose != '-1':
+            if choose == '1':
```
ranbentzvi: should be const

ATM.py
```diff
@@ -0,0 +1,120 @@
...
+            words = line.split()
+            balance_dict[words[0]] = words[2]
+    return balance_dict
+
+
+def check_id(id_input):
+    if id_input not in passwords_dictionary().keys():
+        print('User Not Found')
+        return False
+    else:
```
ranbentzvi: you don't need this else, you can put the `return True` outside the if statement and it would have the same result.

ATM.py
```diff
@@ -0,0 +1,120 @@
+__author__ = 'Adi Kadar Levi'
```
ranbentzvi, line 1: please create an atm class. This class should contain all the data and functions

ranbentzvi (changes requested):

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+class ATM:
+    def __init__(self):
+        self.id_num = 0
+        self.password = 0
+        self.balance = 0
+        self.__pass_dict = {}
+        self.__bal_dict = {}
+
+    def get_passwords_dictionary(self):
+        bank_data_path = r'C:\Users\Elior Mor Yosef\Documents\GitHub' \
```
ranbentzvi: should be const

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+                         r'\python-training\python-training\bank.txt'
+
+        with open(bank_data_path, 'r') as bank_data:
+            for line in bank_data:
+                words = line.split()
+                self.__pass_dict[words[0]] = words[1]
```
ranbentzvi: make the numbers consts

ranbentzvi: and then think of a way for combining the two get functions, they look quite similar.

ATM.py
```diff
@@ -0,0 +1,141 @@
+__author__ = 'Adi Kadar Levi'
+
+
+class ATM:
+    def __init__(self):
+        self.id_num = 0
+        self.password = 0
+        self.balance = 0
+        self.__pass_dict = {}
+        self.__bal_dict = {}
```
ranbentzvi: balance will be better  then bal

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+                         r'\python-training\python-training\bank.txt'
+
+        with open(bank_data_path, 'r') as bank_data:
+            for line in bank_data:
+                words = line.split()
+                self.__bal_dict[words[0]] = words[2]
+        return self.__bal_dict
+
+    def check_id(self, id_input):
+        if id_input not in self.get_passwords_dictionary().keys():
```
ranbentzvi: you don't have to read the data from the file each time...
you already put the data in a dictionary of the class, read the data one time in the contractor.

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+            print('User Not Found')
+            return False
+        return True
+
+    @staticmethod
+    def check_password(user_id, input_password, pass_data):
```
ranbentzvi: use the id data from the self variable, that's why it's there, you should use the class variables that's why you created them.

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+        while pass_data[user_id] != input_password:
```
ranbentzvi: please add an option for user id change here, if I made mistake with it I will be stuck in an infinite loop here

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+            for key in pass_dict:
+                print(key, pass_dict[key], balance_dict[key], file=new_bank)
+        pass
+
+    @staticmethod
+    def set_deposit(user_id, amount_to_deposit, pass_dict,
+                    balance_dict):
+        bank_data_path = r'C:\Users\Elior Mor Yosef\Documents\GitHub' \
+                         r'\python-training\python-training\bank.txt'
+        balance_dict[user_id] = float(amount_to_deposit) + float(balance_dict
```
ranbentzvi: again use the balance variable

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+                                                                 [user_id])
+
+        with open(bank_data_path, 'w') as new_bank:
+            for key in pass_dict:
+                print(key, pass_dict[key], balance_dict[key], file=new_bank)
```
ranbentzvi: please use write\writeline functions for files it's more clear that way

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+        balance_dict[user_id] = float(balance_dict
+                                      [user_id]) - float(amount_to_withdraw)
+
+        with open(bank_data_path, 'w') as new_bank:
+            for key in pass_dict:
+                print(key, pass_dict[key], balance_dict[key], file=new_bank)
+        pass
+
+    @staticmethod
+    def set_deposit(user_id, amount_to_deposit, pass_dict,
```
ranbentzvi: the name should be just deposit without the set (here and in all the other sets)

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+                print(key, pass_dict[key], balance_dict[key], file=new_bank)
+        pass
+
+    @staticmethod
+    def set_change_password(new_password, user_id, pass_dict, balance_dict):
+        bank_data_path = r'C:\Users\Elior Mor Yosef\Documents\GitHub' \
+                         r'\python-training\python-training\bank.txt'
+
+        pass_dict[user_id] = new_password
+        with open(bank_data_path, 'w') as new_bank:
```
ranbentzvi: create a function for data file update

ATM.py
```diff
@@ -0,0 +1,141 @@
...
+            for key in pass_dict:
+                print(key, pass_dict[key], balance_dict[key], file=new_bank)
+        pass
+
+
+def main():
+    user = ATM()
+
+    pass_dict = user.get_passwords_dictionary()
```
ranbentzvi: should be part of the class, and occur automatically in the class init

ATM.py
ranbentzvi: put this loop logic in the class, and use consts. no magic numbers please

ranbentzvi (changes requested):

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
```diff
@@ -1,39 +1,36 @@
...
-    def check_id(self, id_input):
-        if id_input not in self.get_passwords_dictionary().keys():
-            print('User Not Found')
-            return False
-        return True
+                id_numbers = words[0]
+                passwords = words[1]
+                balances = words[2]
+                self.pass_dict[id_numbers] = passwords
+                self.balance_dict[words[0]] = balances
```
ranbentzvi: use id_number here also

ATM.py
```diff
@@ -44,97 +41,81 @@ def check_password(user_id, input_password, pass_data):
...
-                                         bal_dict)
-                print('password successfully changed')
-                break
-            else:
-                print('invalid input')
-                break
+    if id_num in user.pass_dict.keys():
+        user.id_num = id_num
+        user.password = input('please enter your password\n')
+        user.check_password(id_num, user.password, user.pass_dict)
```
ranbentzvi: all the variables for this function are part of the class, why pass them again and not use them as class variables?

ATM.py
```diff
@@ -44,97 +41,81 @@ def check_password(user_id, input_password, pass_data):
...
+
+        balance_choice = '1'
```
ranbentzvi: those consts should be defined like `BANK_DATA_PATH `

ATM.py
```diff
@@ -44,97 +41,81 @@ def check_password(user_id, input_password, pass_data):
...
+        deposit_choice = '2'
+        withdraw_choice = '3'
+        change_pass_choice = '4'
+        exit_choice = '-1'
+        user.choice = user.select()
+        if user.choice == balance_choice:
+            balance = user.get_balance(id_num, user.balance_dict)
```
ranbentzvi: I want you to go over all of your functions and look at their variables.
all the point of making a class variable is that you can use it from the functions in the class, here you keep passing those variables to your functions.

ranbentzvi (changes requested):

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ATM.py
```diff
@@ -30,31 +35,59 @@ def select(self):
...
+        if self.choice == BALANCE_CHOICE:
+            balance = self.balance_dict[self.id_num]
+            print('your balance is {}'.format(balance))
+            print('Thank you for using our services\n'
+                  'For this action '
+                  'you will be charged for 300$ commission\n'
+                  'Bye Bye!')
+
+        elif self.choice == DEPOSIT_CHOICE:
+            amount = float(input('How much would you like to deposit?\n'))
```
ranbentzvi: what would happen if the float convention fails? (in case someone enters a string for example?)

ATM.py
```diff
@@ -30,31 +35,59 @@ def select(self):
                             'To withdraw money press 3\n'
                             'To change password press 4\n'
                             'for Exit press -1\n')
-        return self.choice
+        if self.choice == BALANCE_CHOICE:
```
ranbentzvi, line 38: make it a function and use it in all place you print the user balance

ranbentzvi (changes requested):

ATM.py
AdiKadarLevi: done

ATM.py
AdiKadarLevi: done

ranbentzvi (approved):

PR #119, opened 2020-03-31
Atm exercise

mayareu:


Atm.py
```diff
@@ -0,0 +1,134 @@
...
+        balance = 2
+        with open(self.file_path, 'r') as file:
+            costumers = [i[:i.find('\n')] for i in file.readlines()]
+        costumers = [i.split(',') for i in costumers]
+        for i in costumers:
+            if i[consumer_id] and i[password] and i[balance]:
+                costumer = Costumer(i[consumer_id], i[password], i[balance])
+                _id = i[consumer_id]
+                self.costumers[_id] = costumer
+        print(self.costumers)
```
ranbentzvi, line 37: this is very unsafe

Atm.py
```diff
@@ -0,0 +1,134 @@
...
+        if int(consumer_id) != -1:
+            try:
+                self.consumer = self.costumers[consumer_id.strip()]
+            except:
+                print("No account for the inserted id, please try again")
+                self.get_consumer()
+        else:
+            self.on = False
+
+    def check_password(self):
```
ranbentzvi, line 50: if all the passwords were bad you don't return anything, it would work with your if in start action function but it's a bad habit. please return something in all cases

Atm.py
```diff
@@ -0,0 +1,134 @@
...
+            if entered_password == self.consumer.password:
+                return True
+                break
+            else:
+                print("The inserted password was wrong")
+
+    def update_details(self):
+        lines = []
+        for consumer in self.costumers.values():
+            lines.append(consumer.__str__())
```
ranbentzvi, line 63: you can do `str(consumer)` instead

Atm.py
```diff
@@ -0,0 +1,134 @@
...
+        with open(self.file_path, 'w') as set_new_file:
+            set_new_file.writelines(lines)
+
+    def get_desired_operation(self):
+        check_balance = 1
```
ranbentzvi, line 68: make those real consts

ranbentzvi (changes requested):

PR #120, opened 2020-03-31
Feature/matrix

kerenGuez:


advanced/matrix/matrix.py
```diff
@@ -0,0 +1,158 @@
...
+        """Create a unity tuple.
+
+        Args:
+            size (int): the number of rows and columns for new matrix.
+
+        Returns:
+            tuple: New tuple of tuples, matrix, main diagonal is 1, the rest 0.
+        """
+        unity_matrix = []
+        for row_num, row in enumerate(range(size)):
```
natsala13: Using enumerate is great also here it have no sense since row and row_num will be identical.

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,158 @@
...
+            tuple_nums = []
+
+            for column in range(size):
+                if column == row_num:
+                    tuple_nums.append(1)
+
+                else:
+                    tuple_nums.append(0)
+
+            unity_matrix.append(tuple(tuple_nums))
```
natsala13: Replace it with a smart list comprehension

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,158 @@
...
+    def ones(cls, size):
+        """Create a new matrix in the given size filled with 1.
+
+        Args:
+            size: The size of the new matrix.
+
+        Returns:
+            tuple: Tuple of tuples filled with ones.
+        """
+        ones_mat = [tuple([1 for num in range(size)]) for i in range(size)]
```
natsala13: Try and use list concatenation for creating list of ones.

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,158 @@
...
+            raise ZeroDivisionError("Cannot divide by 0.")
+        return Matrix(tuple((tuple((num / other for num in each_tuple))
+                             for each_tuple in self.tuples)))
+
+    def __add__(self, other):
+        """Return result of 2 matrices addition."""
+        return Matrix(tuple((tuple((x + y for x, y in zip(tuple1, tuple2))))
+                      for tuple1, tuple2 in zip(self.tuples, other)))
+
+    def __sub__(self, other):
```
natsala13, line 189: Reuse already implemented methods if you can.

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,158 @@
...
+            return self._multiplication(other.tuples)
+
+        else:
+            raise TypeError("Second parameter must be int/matrix.")
+
+    def __rmul__(self, other):
+        """Return the regular multiplication when the order is reversed."""
+        return self * other
+
+    def __truediv__(self, other):
```
natsala13, line 158: Reuse already implemented methods if you can.

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,158 @@
...
+
+            Args:
+                sec_mat (tuple): A matrix to multiply with.
+
+            Returns:
+                tuple: Result matrix of the multiplication between 2 matrices.
+        """
+        result = [tuple(sum(n_a * n_b for n_a, n_b in zip(a_row, b_col))
+                        for b_col in zip(*sec_mat))
+                  for a_row in self.tuples]
```
natsala13: Excellent !!!

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,158 @@
...
+        return tuple(result)
+
+    def _multiply_by_scalar(self, scalar):
+        """Return new matrix multiplied by scalar."""
+        return Matrix(tuple((tuple((num * scalar for num in each_tuple))
+                             for each_tuple in self.tuples)))
```
natsala13: Don't need self.tuples
Why? (understand it before changing).

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,158 @@
...
+            Matrix: The result matrix.
+        """
+        if isinstance(other, int):
+            return self._multiply_by_scalar(other)
+
+        elif isinstance(other, Matrix):
+            return self._multiplication(other.tuples)
+
+        else:
+            raise TypeError("Second parameter must be int/matrix.")
```
natsala13: I think argument would fir here better.
Who is the second? will __mul__ be called only when a matrix is at the left?

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,158 @@
...
+        """Return new matrix divided by scalar."""
+        if other == 0:
+            raise ZeroDivisionError("Cannot divide by 0.")
+        return Matrix(tuple((tuple((num / other for num in each_tuple))
+                             for each_tuple in self.tuples)))
+
+    def __add__(self, other):
+        """Return result of 2 matrices addition."""
+        return Matrix(tuple((tuple((x + y for x, y in zip(tuple1, tuple2))))
+                      for tuple1, tuple2 in zip(self.tuples, other)))
```
natsala13: Correct indentation.

natsala13: Way to much brackets.

natsala13 (changes requested):

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
+class Matrix:
```
rfire01, line 4: python file should start with doc line.

rfire01, line 4: Add unittests

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+
+        else:
+            raise TypeError("Object is not a tuple's matrix.")
+
+    def is_good_matrix(self):
+        """Check that matrix given is valid."""
+        self._is_tuple()
+        self._is_even()
+
+    def __init__(self, matrix_tuples):
```
rfire01: init should be the first function in class

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
+class Matrix:
+    """A class used to represent an immutable n*n matrix.
+
+        Attributes:
+            __matrix (tuple): a tuple of tuples representing a matrix.
+    """
+    def _is_even(self):
```
rfire01: The name of the function make it sound that the function return true\false if its even

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+        """Check that matrix has equal number of rows and columns."""
+        for mat_tuple in self.__matrix:
+            if len(mat_tuple) != len(self.__matrix):
+                raise ValueError("The matrix is not n*n")
+
+    def _is_tuple(self):
+        """Check that matrix is made of a tuple of tuples."""
+        if isinstance(self.__matrix, tuple):
+            for mat_tuple in self.__matrix:
+                if not isinstance(mat_tuple, tuple):
```
rfire01: can you think about a way to do it with list comprehension?

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+    def unity(cls, size):
+        """Create a unity tuple.
+
+        Args:
+            size (int): the number of rows and columns for new matrix.
+
+        Returns:
+            tuple: New tuple of tuples, matrix, main diagonal is 1, the rest 0.
+        """
+        lis = [tuple(1 if col == row else 0 for col in range(size))
```
rfire01: Why not make it a tuple from the start instead of list?

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+            tuple: Tuple of tuples filled with ones.
+        """
+        ones_mat = (((1,) * size,) * size)
+        return cls(ones_mat)
+
+    def _multiplication(self, sec_mat):
+        """Calculate the multiplication of 2 matrices.
+
+            Args:
+                sec_mat (tuple): A matrix to multiply with.
```
rfire01: give full names

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+
+            Returns:
+                tuple: Result matrix of the multiplication between 2 matrices.
+        """
+        result = [tuple(sum(n_a * n_b for n_a, n_b in zip(a_row, b_col))
+                        for b_col in zip(*sec_mat))
+                  for a_row in self]
```
rfire01: Very good!
just save the conversion from list to tuple

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+        return tuple(result)
+
+    def _multiply_by_scalar(self, scalar):
+        """Return new matrix multiplied by scalar."""
```
rfire01: missing doc

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+        return Matrix(tuple((tuple((num * scalar for num in each_tuple))
```
rfire01: Too many parenthesis

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+        """Multiply matrix by scalar or another matrix.
+
+        Args:
+            other (int/Matrix): Scalar or another matrix.
+
+        Returns:
+            Matrix: The result matrix.
+        """
+        if isinstance(other, Matrix):
+            return self._multiplication(other.tuples)
```
rfire01: blank line

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+                        for b_col in zip(*sec_mat))
+                  for a_row in self]
+        return tuple(result)
+
+    def _multiply_by_scalar(self, scalar):
+        """Return new matrix multiplied by scalar."""
+        return Matrix(tuple((tuple((num * scalar for num in each_tuple))
+                             for each_tuple in self)))
+
+    def __mul__(self, other):
```
rfire01, line 118: Will this function always work? for any two matrices?

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Args:
+            other (int/Matrix): Scalar or another matrix.
+
+        Returns:
+            Matrix: The result matrix.
+        """
+        if isinstance(other, Matrix):
+            return self._multiplication(other.tuples)
+        return self._multiply_by_scalar(other)
```
rfire01: what will happen if I do Matrix() * [1,2,3]?

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __rmul__(self, other):
+        """Return the regular multiplication when order is reversed"""
+        return self * other
```
rfire01, line 99: This is not correct

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __truediv__(self, other):
+        """Return new matrix divided by scalar."""
+        if other == 0:
+            raise ZeroDivisionError("Cannot divide by 0.")
```
rfire01: blank line

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+        return self * (1 / other)
+
+    def __add__(self, other):
+        """Return result of 2 matrices addition."""
+        add_result = ((tuple((x + y for x, y in zip(tuple1, tuple2))))
+                      for tuple1, tuple2 in zip(self, other))
```
rfire01: too many parenthesis

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+        """Return result of 2 matrices subtraction."""
+        return self + (-1 * other)
+
+    def __iter__(self):
+        """Return an iterator on the matrix"""
+        return iter(self.tuples)
+
+    def __eq__(self, other):
+        """Check for equality between matrices."""
+        return self.tuples == other
```
rfire01: does it work?

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __ne__(self, other):
+        """Check if the two matrices are not equal."""
+        return self.tuples != other
```
rfire01: same

rfire01 (changes requested):

advanced/matrix/matrix.py
```diff
@@ -72,78 +64,106 @@ def ones(cls, size):
...
 
     def __eq__(self, other):
         """Check for equality between matrices."""
-        return self.tuples == other
+        return self.tuples == other.tuples
 
     def __ne__(self, other):
         """Check if the two matrices are not equal."""
-        return self.tuples != other
+        return self.tuples != other.tuples
```
natsala13: Try reusing already implemented method.

natsala13 (changes requested):

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,120 @@
...
+    def test_initialize(self):
+        """Test initialization of matrix."""
+        matrix_tuples = ((2, 5), (8, 3))
+        my_matrix = Matrix(matrix_tuples)
+
+        self.assertEqual(my_matrix.tuples, matrix_tuples)
+
+        with self.assertRaises(TypeError):
+            Matrix("hi")
+            Matrix(((1, 2, 3), [4, 5, 6], 5))
```
natsala13, line 22: This line is unreachable

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def setUp(self):
+        """Initialize matrices a and b before tests."""
+        self.mat_a = Matrix(((1, 2), (3, 4)))
+        self.mat_b = Matrix(((3, 5), (6, 8)))
+
+    def test_initialize(self):
+        """Test initialization of matrix."""
+        matrix_tuples = ((2, 5), (8, 3))
+        my_matrix = Matrix(matrix_tuples)
```
natsala13: You have already initialized a matrix in setup, why creating another one?

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,120 @@
...
+
+        with self.assertRaises(ValueError):
+            self.mat_a - Matrix.ones(3)
+
+    def test_iter(self):
+        """Test the iterator of the matrix."""
+        lines = ""
+        for line in self.mat_a:
+            lines += f"{line}\n"
+        self.assertEqual(lines, "(1, 2)\n(3, 4)\n")
```
natsala13: Instead of using string just check every line of the matrix

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,120 @@
...
+        """Test the equality check between matrices."""
+        self.assertEqual(self.mat_a == self.mat_b, False)
+
+    def test_not_equal(self):
+        """Test if two matrices are not equal."""
+        self.assertEqual(self.mat_a != self.mat_b, True)
+
+    def test_hash(self):
+        """Test if hash method."""
+        dictionary = {Matrix(((1, 1), (2, 2))): 2,
```
natsala13: You already initialized matrices in setup method.

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,120 @@
+"""Unit testing for the Matrix class."""
```
natsala13, line 1: **Change all methods so that every test case would check only one case. (one assert). If you want to check a few test cases of add_matrix for example, then write few different unit test methods checking every case separately.**

natsala13 (changes requested):

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+            Args:
+                other_matrix (tuple): A matrix to multiply with.
+
+            Returns:
+                tuple: Result matrix of the multiplication between 2 matrices.
+        """
+        result = tuple(tuple(sum(n_a * n_b for n_a, n_b in zip(a_row, b_col))
+                       for b_col in zip(*other_matrix))
+                       for a_row in self)
+        return Matrix(result)
```
natsala13: If one want to create a new class inheriting from this one, and try to use this method from the inheriting class. What would happen?

Anyway, try not using Hard Coded class name.
We are looking for a code which require a minimum amount of work for every adaptation we want to add to it.

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
+"""An implementation of the class for the data type 'Matrix'."""
```
natsala13, line 1: Great Work!

natsala13 (changes requested):

natsala13 (approved):

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+        """Check that matrix has equal number of rows and columns."""
+        for mat_tuple in self.__matrix:
+            if len(mat_tuple) != len(self.__matrix):
+                raise ValueError("The matrix is not n*n")
+
+    def _check_tuple(self):
+        """Check that matrix is made of a tuple of tuples."""
+        if isinstance(self.__matrix, tuple):
+            if not all(isinstance(mat_tuple, tuple)
+                       for mat_tuple in self.__matrix):
```
rfire01: You should use naming that will be as clear as possible for other people who don't know your code.
For example, instead of mat_tuple, maybe matrix_row (to make it clear what this tuple means)

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+
+                raise TypeError("Object is not a tuple's matrix.")
```
rfire01: Object is not tuple of tuples

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def _multiply_by_scalar(self, scalar):
+        """Return new matrix multiplied by scalar.
+
+        Args:
+            scalar (int/float): a number to multiply matrix with.
+
+        Returns:
+            Matrix : The new matrix multiplied by scalar.
+
```
rfire01: remove

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+        Args:
+            size: The size of the new matrix.
+
+        Returns:
+            tuple: Tuple of tuples filled with ones.
+        """
+        ones_mat = (((1,) * size,) * size)
+        return cls(ones_mat)
+
+    def _multiplication(self, other_matrix):
```
rfire01: if you have several multiplication, be specific in the function name

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+            other (float/int/Matrix): Scalar or another matrix.
+
+        Returns:
+            Matrix: The result matrix.
+        """
+        try:
+            self._check_valid_matrix(other)
+            return self._multiplication(other.tuples)
+
+        except TypeError:
```
rfire01: There is no need for try-except here

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+                raise TypeError("Can only multiply numbers and matrices.")
+
+    def _check_valid_matrix(self, other):
+        """Check that 'other' is a matrix of the same size as self."""
+        if isinstance(other, Matrix):
+            if len(other.tuples) != len(self.tuples):
+                raise ValueError("Cannot operate on different sized matrices.")
+
+        else:
+            raise TypeError("Cannot operate on different sized matrices.")
```
rfire01, line 152: This doesn't seem correct

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def __rmul__(self, other):
+        """Return the regular multiplication when in reversed order."""
+        return self * other
```
rfire01, line 156: If other is matrix this is not correct.
Can you tell me why even so, this answer is still correct?

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def __truediv__(self, other):
```
rfire01, line 158: Fix in all functions:
When there is an option that you will raise an exception in function, you need in the function doc add after the "returns" a section of "Raises" and what can it raise (you can look at the coding style doc how it looks)

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,178 @@
...
+            return self * (1 / other)
+
+        else:
+            raise TypeError("'other' argument must be a number.")
+
+    def __add__(self, other):
+        """Return result of 2 matrices addition."""
+        self._check_valid_matrix(other)
+        add_result = (tuple(x + y for x, y in zip(tuple1, tuple2))
+                      for tuple1, tuple2 in zip(self, other))
```
rfire01: same, better naming

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,148 @@
...
+            Matrix(((1, 2, 3), (1, 2, 3)))
+
+    def test_unity_2_dimensional(self):
+        """Test unity method on 2 by 2 matrix."""
+        self.assertEqual(Matrix.unity(2), Matrix(((1, 0), (0, 1))))
+
+    def test_unity_3_dimensional(self):
+        """Test unity method on 3 by 3 matrix."""
+        self.assertEqual(Matrix.unity(3),
+                         Matrix(((1, 0, 0), (0, 1, 0), (0, 0, 1))))
```
rfire01: You may missed it, but you checked here two functions:
1. unity function
2. __eq__ function
Generally it is ok to use multiple functions in single test, but you need to check they are working before using them (the functions that are not the purpose of the test). 
In this case the test might have failed, but because of the __eq__ and not because the unity
(This is correct for all tests that do it)

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def test_multiplication_by_float(self):
+        """Test multiplication by float."""
+        self.assertEqual(self.mat_a * 0.9, Matrix(((0.9, 1.8), (2.7, 3.6))))
+
+    def test_multiplication_of_matrices(self):
+        """Test multiplication of matrices."""
+        self.assertEqual(self.mat_a * self.mat_b, Matrix(((15, 21), (33, 47))))
+
+    def test_multiplication_reversed(self):
```
rfire01: what about matrix mul reversed?

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,148 @@
...
+        self.assertEqual(self.mat_a / 0.1, Matrix(((10, 20), (30, 40))))
+
+    def test_division_Type_validation(self):
+        """Test that division is only allowed with a number."""
+        with self.assertRaises(TypeError):
+            self.mat_a / self.mat_b
+
+    def test_addition(self):
+        """Test the method for addition of 2 matrices."""
+        self.assertEqual(self.mat_a + Matrix.unity(2),
```
rfire01: why not used mat_b if you already created it?

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,148 @@
...
+        """Test that you can subtract only same sized matrices."""
+        with self.assertRaises(ValueError):
+            self.mat_a - Matrix.ones(3)
+
+    def test_iter(self):
+        """Test the iterator of the matrix."""
+        mat_a_iter = iter(self.mat_a)
+        self.assertEqual(next(mat_a_iter), (1, 2))
+
+        self.assertEqual(next(mat_a_iter), (3, 4))
```
rfire01: instead of using next and splitting to 2 asserts, you can do something like list(mat_a) or tuple(mat_a) etc., and then compare the resulted list\tuple. 
list and tuple know how to iterate on iterable variable

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def test_eq(self):
+        """Test the equality check between matrices."""
+        self.assertEqual(self.mat_a == self.mat_b, False)
```
rfire01: This is not enough.
This test shows that comparing two matrices will return False, however, this test will pass even if the __eq__ function was
def __eq__(...):
   return False
When testing you need to make sure all of its functionality works

rfire01 (changes requested):

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,220 @@
...
+
+class Matrix:
+    """A class used to represent an immutable n*n matrix.
+
+        Attributes:
+            matrix_rows (tuple): a tuple of tuples representing a matrix.
+
+        Raises:
+            TypeError: self is not a tuple of tuples.
+            ValueError: self doesn't have an equal number of rows and columns.
```
rfire01: self doesn't exists yet at this point (you should be explicit what are you talking about)

advanced/matrix/matrix.py
```diff
@@ -0,0 +1,220 @@
...
+            raise TypeError("Cannot operate on different sized matrices.")
+
+    def __rmul__(self, other):
+        """Return the regular multiplication when in reversed order."""
+        return self * other
+
+    def __truediv__(self, other):
+        """Return new matrix divided by scalar.
+        Args:
+            other (float/int): Scalar to divide by.
```
rfire01: in those cases you can just write number as the type

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,166 @@
...
+    def test_tuples_property(self):
+        """Test tuples property of matrix."""
+        self.assertEqual(self.mat_a.matrix_rows, ((1, 2), (3, 4)))
+
+    def test_type_validation(self):
+        """Test that only a tuple of tuples is allowed in Matrix."""
+        with self.assertRaises(TypeError):
+            Matrix(((1, 2, 3), [4, 5, 6], 5))
+
+    def test_value_validation(self):
```
rfire01: maybe test_invalid_constructor or something like this.
its not clear from the name what you are testing

advanced/matrix/test_matrix.py
```diff
@@ -0,0 +1,166 @@
...
+    def test_repr(self):
+        """Test matrix instance representation."""
+        self.assertEqual(self.mat_a.__repr__(), "Matrix(((1, 2), (3, 4)))")
+
+    def test_str(self):
+        """Test textual representation of matrix instance."""
+        self.assertEqual(str(self.mat_a), "((1, 2), (3, 4))")
+
+
+if __name__ == "__main__":
```
rfire01, line 165: missing test for tuples property

rfire01 (changes requested):

PR #124, opened 2020-04-01
Added a python style guide

jhb1996:
Style guide for commenting in python


rfire01:
Fix the file format so it will be more readable (like the readme file)

PR #127, opened 2020-04-01
Matrix exercise

mayareu:


Matrix.py
```diff
@@ -0,0 +1,140 @@
...
+            for column in xrange(size):
+                member.append(1)
+            ones_matrix.append(tuple(member))
+        self.__init__(tuple(ones_matrix))
+
+    def tupels(self):
+        return self.__matrix
+
+    def scalar_multiplication(self, scalar):
+        result = [tuple(map(lambda x: x * scalar, i)) for i in self.__matrix]
```
ranbentzvi: the use of map and lambda is not recommended, please change that

Matrix.py
```diff
@@ -0,0 +1,140 @@
...
+        return tuple(result)
+
+    def addition(self, other_matrix):
```
ranbentzvi, line 70: your addition and subtraction look quite similar, think of a way to unify them

Matrix.py
```diff
@@ -0,0 +1,140 @@
...
+    def size_comparision(self, matrix):
+        if self.get_size() == matrix.get_size():
+            return True
+        else:
+            return False
+
+    def get_size(self):
+        return self.__matrix.__len__()
+
+    def compare(self, other_matrix):
```
ranbentzvi, line 121: if the other object type is not matrix?

ranbentzvi (changes requested):

PR #128, opened 2020-04-01
first draft of 'atm'

shiravak:


exercises/ex_atm.py
```diff
@@ -83,8 +84,9 @@ def read_dict_move_to_file(file_path, atm_dict):
     """Move the updated dictionary to the atm information file.
 
     Args:
-    file_path (string): path to atm information file.
-    atm_dict (dictionary): the dictionary, containing all the atm information.
+        file_path (string): path to atm information file.
+        atm_dict (dictionary): the dictionary, containing all the atm
+                               information.
```
jhb1996: indent for a continued line should only be 4 spaces inward from the previous line.

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+                line_list = line.rstrip().split(',')
+                atm_options_dict.update({line_list[0]: [line_list[1],
+                                                        float(line_list[2])]})
+        return atm_options_dict
+
+    except FileNotFoundError:
+        print("The path to the file you have entered was incorrect.")
+
+    except Exception as e:
+        print(e)
```
IgalKolihman: is there a reason that you catch just any exception here and print it?
if no, it is a **VERY BAD** habit to except the exception of the class `Exception`.

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
+# ATM Exercise
```
IgalKolihman: it's very good that you added documentation for the script.
but when you do that, use `"""My documentation"""` and not `# My documentation`

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+
+def read_file_move_to_dict(file_path):
+    """Read file and move atm information to dictionary.
+
+    Args:
+        file_path (string): path to atm information file.
+
+    Return:
+        A dictionary, the key is each users id.
+        the value is a list of user's password and balance.
```
IgalKolihman: In the `Return:` section of the documentation, the doc need to start with the type of the returned value, and then the description of that value.

```suggestion
       dict. With users ID as keys and user's data as values.
```

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
+# ATM Exercise
+
+import sys
+
+FILE_PATH = 1
+PASSWORD_INDEX = 0
+BALANCE_INDEX = 1
+
+
+def read_file_move_to_dict(file_path):
```
IgalKolihman: bad naming

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+    """Read file and move atm information to dictionary.
+
```
IgalKolihman, line 11: the actions of this function look a bit complicated, and it's hard to understand what you are doing.
Add a description here of what you expect the file to look like.

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+    Args:
+        file_path (string): path to atm information file.
+
+    Return:
+        A dictionary, the key is each users id.
```
jhb1996: write the return as follows:

Returns:
       <type>: information about it.

See here for most of the important style things you need to know https://github.com/osherdp/python-training/blob/yaakov_branch/Style_Guide.md

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+FILE_PATH = 1
+PASSWORD_INDEX = 0
+BALANCE_INDEX = 1
+
+
+def read_file_move_to_dict(file_path):
+    """Read file and move atm information to dictionary.
+
+    Args:
+        file_path (string): path to atm information file.
```
IgalKolihman: in the brackets, you need to specify the type of the attribute.
in python, the type of string is `str`, not `string`.

```suggestion
        file_path (str): path to atm information file.
```

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+
+    except FileNotFoundError:
+        print("The path to the file you have entered was incorrect.")
+
+    except Exception as e:
+        print(e)
+
+
+def check_input_execute(atm_dict, user_input, user_id):
+    """Check user option input, and act accordingly, if input is correct.
```
IgalKolihman:
```suggestion
    """Check user option input, and act accordingly if the input is correct.
```

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Args:
+        atm_dict (dictionary): the dictionary, containing all the atm
```
IgalKolihman: in python, the type of dictionary is `dict`, not `dictionary`.

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+        return atm_options_dict
+
+    except FileNotFoundError:
+        print("The path to the file you have entered was incorrect.")
+
+    except Exception as e:
+        print(e)
+
+
+def check_input_execute(atm_dict, user_input, user_id):
```
IgalKolihman: this function is too long. 
separate it to smaller, more action specified, functions
for example: 
- function for withdrawal
- function for deposit
...

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+            print("Password changed to: " + new_password)
+
+        else:
+            print("Invalid input")
+
+    except ValueError:
+        print("Invalid input")
+
+    except Exception as e:
+        print(e)
```
IgalKolihman: never do that

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Args:
+        file_path (string): path to atm information file.
+
+    Return:
+        A dictionary, the key is each users id.
+        the value is a list of user's password and balance.
+    """
+    try:
+        atm_options_dict = {}
```
jhb1996: We don't use the type in the variable name. You can just call it atm_options.

Look over your code and change the name of all of your variable names which contain types (pycharm's refactor function should make it easier).

exercises/ex_atm.py
IgalKolihman: Your function names in the code could look better. There are some functions that, as I understood, didn't match the documentation.

Please address it.

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+        atm_dict (dictionary): the dictionary, containing all the atm
+                               information.
+        user_input (string): the option choice of the user.
+        user_id (string): the id of the user, to find him in the dictionary.
+
+    Return:
+        The new changed dictionary of atm information.
+    """
+    try:
+        # ['check', 'withdrawal', 'deposit', 'change']:
```
IgalKolihman: this should also be in the function documentation

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+            for key in atm_dict:
+                file_pointer.write(','.join([str(key), atm_dict[key][0],
+                                             str(atm_dict[key][1])]) + "\n")
+        print("Changes saved :)")
+
+    except FileNotFoundError:
+        print("Something went wrong, when accessing file")
+
+    except Exception as e:
+        print(e)
```
IgalKolihman: bad habit

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+    while True:
+        user_id_input = input("Enter costumer ID, to turn off the ATM enter "
+                              "-1 -> ")
+        if user_id_input == -1:
+            atm_options_dict = None
+            break
+        elif user_id_input in atm_options_dict:
+            break
+        else:
+            print("Invalid ID")
```
IgalKolihman: make this a function

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+
+    while True and atm_options_dict is not None:
```
IgalKolihman: as i see it, there is no need for the `True and`

```suggestion
    while atm_options_dict is not None:
```

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+        if user_option == "-1":
+            read_dict_move_to_file(file_path, atm_options_dict)
+            break
+        elif user_option == "options":
+            print("The ATM allow the following options to a using costumer:" +
+                  "\n1. 'check' the balance.\n2. cash 'withdrawal'." +
+                  "\n3. cash 'deposit'.\n4. 'change' password.")
+        else:
+            atm_options_dict = check_input_execute(atm_options_dict,
+                                                   user_option, user_id_input)
```
IgalKolihman: make this a function

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def main():
+    file_path = sys.argv[FILE_PATH]
+    atm_options_dict = read_file_move_to_dict(file_path)
+
+    while True:
+        user_id_input = input("Enter costumer ID, to turn off the ATM enter "
+                              "-1 -> ")
+        if user_id_input == -1:
```
jhb1996: You need to convert the input to a number first in order for this if statement to work

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+        elif user_id_input in atm_options_dict:
+            break
+        else:
+            print("Invalid ID")
+
+    while True and atm_options_dict is not None:
+        user_option = input("Enter wanted option, to turn off the ATM enter -1"
+                            + "\nEnter 'options' to see ATM's options. -> ")
+        user_option = user_option.lower()
+        if user_option == "-1":
```
IgalKolihman: it looks like you are going to use the `-1` multiple times in your code.
when you have such a case, better save the `-1` into a CONSTANT variable, and use the variable instead of the `-1` in your code.
this way, if some changes are about to come (for example, `-1` changed to `"turn off"`), it will be easy for you to handle it.

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+        if user_id_input == -1:
+            atm_options_dict = None
+            break
+        elif user_id_input in atm_options_dict:
+            break
+        else:
+            print("Invalid ID")
+
+    while True and atm_options_dict is not None:
+        user_option = input("Enter wanted option, to turn off the ATM enter -1"
```
jhb1996: You don't need the ''+" 

When a string goes onto the next line python understands that its one big string (even though each line has its own quotation marks)

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
+# ATM Exercise
+
+import sys
+
+FILE_PATH = 1
```
IgalKolihman: this constant variable sounds like a string. why is it equal to 1?
maybe add a comment that describes these constant variables

IgalKolihman (changes requested):

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Args:
+        file_path (string): path to atm information file.
+        atm_dict (dictionary): the dictionary, containing all the atm
+                               information.
+    """
+    try:
+        with open(file_path, 'w') as file_pointer:
+            for key in atm_dict:
+                file_pointer.write(','.join([str(key), atm_dict[key][0],
```
jhb1996: This will be easier to read if you first assign atm_dict[key][0] and atm_dict[key][1] to variables. That way it will be clear exactly what you are writing to the file without someone going back to you dictionary and seeing how values are stored inside it.

exercises/ex_atm.py
jhb1996: good use of constants.

However is this case the name is a but unclear. FILE_PATH sounds like the string telling me where the file goes not the index of the file path in the arguments

exercises/ex_atm.py
jhb1996: For cases like this we usually use tuples instead of lists. Tuples are immutable (cannot be overwritten) so we use them for cases where we aren't changing (ie. mutating) the information inside of them.

All you will need to do to change the brackets you used to create the list to parenthesis to create a tuple.

read more here especially number 11. https://data-flair.training/blogs/python-tuples-vs-lists/

jhb1996 (commented):
good start.

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+        elif user_input == "deposit":
+            deposit = input("Enter how much you want to withdrawal -> ")
+            if float(deposit) > 0:
+                atm_dict[user_id][BALANCE_INDEX] += float(deposit)
+                print("Deposit succeeded")
+
+        elif user_input == "change":
+            new_password = input("Enter you'r new password -> ")
+            atm_dict[user_id][PASSWORD_INDEX] = new_password
+            print("Password changed to: " + new_password)
```
natsala13: Your function is too long, divide it to smaller methods.
It will make your code clearer and more readable even if it will be a bit longer.

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    try:
+        with open(file_path, 'w') as file_pointer:
+            for key in atm_dict:
+                file_pointer.write(','.join([str(key), atm_dict[key][0],
+                                             str(atm_dict[key][1])]) + "\n")
+        print("Changes saved :)")
+
+    except FileNotFoundError:
+        print("Something went wrong, when accessing file")
```
natsala13: The program is raising an exception saying the file was not found.
You catch it and instead print only partial information that something went wrong.
First thing would be to print all the information about the file the user tried to open and the fact that it doesn't exist.
Or you could simply just let the program crash throwing an exception saying that the file was not found.
Chose wisely ;)

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+        elif user_input == "change":
+            new_password = input("Enter you'r new password -> ")
+            atm_dict[user_id][PASSWORD_INDEX] = new_password
+            print("Password changed to: " + new_password)
+
+        else:
+            print("Invalid input")
+
+    except ValueError:
+        print("Invalid input")
```
natsala13, line 92: When will that happen? I see that if a user input a non existing command that would be noticed above.

exercises/ex_atm.py
```diff
@@ -0,0 +1,137 @@
...
+        The new changed dictionary of atm information.
+    """
+    try:
+        # ['check', 'withdrawal', 'deposit', 'change']:
+        if user_input == "check":
+            print(f"You'r balance: {atm_dict[user_id][BALANCE_INDEX]}")
+
+        elif user_input == "withdrawal":
+            withdrawal = input("Enter how much you want to withdrawal -> ")
+            if float(withdrawal) > 0:
```
natsala13: If a user entered a negative amount, nothing will happen and the user will not even know it. output something to the user telling him his input wasn't correct.

natsala13 (changes requested):

exercises/ex_atm.py
shiravak: but i need to change the vlues in it, when the user want to change password or deposit or withdraw money. i think that if i will use a tuple, i will need to cahnge it to list, change the thing the user want and then change it back.

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
+"""ATM Exercise"""
+
```
rfire01: remove

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+import sys
+
+FILE_PATH_INDEX = 1  # describe placement of the argument containing file path.
+PASSWORD_INDEX = 0
+BALANCE_INDEX = 1
+END_ATM_SERVICE = "-1"
```
rfire01: reorder

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def store_file_content(file_path):
+    """Read file and move atm information to dictionary.
+    The lines of the file will contain user's id, balance and password,
```
rfire01: blank line between function description and additional notes

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+    with comma's separating them.
+
+    Args:
+        file_path (str): path to atm information file.
+
+    Return:
+        dict. with user's id as keys,
+        and user's password and balance in a list as values.
```
rfire01: if you have more then 1 line for return \ arg doc, the lines after the first one should start with indent
for example:
dict. something
____continue something (it doesn't show space in comment, so I putted _ instead)

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+            for line in file_pointer:
+                line_list = line.rstrip().split(',')
+                atm_options.update({line_list[0]: [line_list[1],
+                                                   float(line_list[2])]})
+
+        return atm_options
+
+    except FileNotFoundError:
+        print(f"Can't access the path to the file you have entered:"
+              f"{file_path}")
```
rfire01: its look like file not found error, not file cannot be accessed

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+        dict. with user's id as keys,
+        and user's password and balance in a list as values.
+    """
+    try:
+        atm_options = {}
+        with open(file_path, 'r') as file_pointer:
+            for line in file_pointer:
+                line_list = line.rstrip().split(',')
+                atm_options.update({line_list[0]: [line_list[1],
+                                                   float(line_list[2])]})
```
rfire01: don't use update for this

rfire01: don't save the information positional (by index).
What happen if for some reason you change the order here?
you need to change every other place you use it, and if you miss 1, your code doesn't work anymore

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+
+        return atm_options
+
+    except FileNotFoundError:
+        print(f"Can't access the path to the file you have entered:"
+              f"{file_path}")
+
+
+def get_user_id(atm_info):
```
rfire01, line 37: doc?

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+    while True:
+        user_id_input = input("Enter costumer ID, to turn off the ATM enter "
+                              "-1 -> ")
+        if user_id_input == END_ATM_SERVICE:
+            atm_info = None
+            return user_id_input, atm_info
+        elif user_id_input in atm_info:
+            return user_id_input, atm_info
+        else:
+            print("Invalid ID")
```
rfire01: blank line between logic blocks (check the entire code)

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
+"""ATM Exercise"""
+
+import sys
+
+FILE_PATH_INDEX = 1  # describe placement of the argument containing file path.
```
rfire01: no need for the doc

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+        print(f"Can't access the path to the file you have entered:"
+              f"{file_path}")
+
+
+def get_user_id(atm_info):
+    while True:
+        user_id_input = input("Enter costumer ID, to turn off the ATM enter "
+                              "-1 -> ")
+        if user_id_input == END_ATM_SERVICE:
+            atm_info = None
```
rfire01: what?

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+            return user_id_input, atm_info
+        elif user_id_input in atm_info:
+            return user_id_input, atm_info
+        else:
+            print("Invalid ID")
+
+
+def atm_service(file_path, atm_options, user_id_input):
```
rfire01, line 49: add doc to all functions

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def check_input(atm_info, user_input, user_id):
+    """Check user option input, and act accordingly if the input is correct.
+
+    Args:
+        atm_info (dict): the dictionary, containing all the atm information.
+        user_input (str): the option choice of the user.
+                          there are four options- check, withdrawal,
+                          deposit and change
```
rfire01: single indent between lines

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+        user_id (str): the id of the user, to find him in the dictionary.
+
+    Return:
+        The new changed dictionary of atm information.
```
rfire01: type?

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+    """Check user option input, and act accordingly if the input is correct.
+
+    Args:
+        atm_info (dict): the dictionary, containing all the atm information.
+        user_input (str): the option choice of the user.
+                          there are four options- check, withdrawal,
+                          deposit and change
+        user_id (str): the id of the user, to find him in the dictionary.
+
+    Return:
```
rfire01: Returns

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+        print(f"You'r balance: {atm_info[user_id][BALANCE_INDEX]}")
+
+    elif user_input == "withdrawal":
+        atm_info = withdrawal(atm_info, user_id)
+
+    elif user_input == "deposit":
+        atm_info = deposit(atm_info, user_id)
+
+    elif user_input == "change":
+        check_balance(atm_info, user_id)
```
rfire01: the user input should be 1-4, not a string

rfire01: check balance? wrong command

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+
+    else:
+        print("Invalid input")
+
+    return atm_info
+
+
+def withdrawal(atm_info, user_id):
+    """function to withdraw money from atm dict"""
```
rfire01: fix doc

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+    withdraw = input("Enter how much you want to withdrawal -> ")
+    if float(withdraw) > 0:
+        atm_info[user_id][BALANCE_INDEX] -= float(withdraw)
```
rfire01: what if I have 0$ in my balance?

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+    else:
+        print("invalid deposit input")
+    return atm_info
+
+
+def check_balance(atm_info, user_id):
+    """function to check user's balance in the atm dict"""
+    new_password = input("Enter you'r new password -> ")
+    atm_info[user_id][PASSWORD_INDEX] = new_password
+    print("Password changed to: " + new_password)
```
rfire01: function name and doc doesn't fit the logic

rfire01: Who would want his password to print on screen :)

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+              f"{file_path}")
+
+
+def main():
+    file_path = sys.argv[FILE_PATH_INDEX]
+    atm_options = store_file_content(file_path)
+
+    user_id_input, atm_options = get_user_id(atm_options)
+
+    while atm_options is not None:
```
rfire01: This doesn't make very much sense.
I would expect to keep waiting for input until -1 received, not until my DB is None
(its counter intuitive)

rfire01: The ATM should support multiple users in one run

exercises/ex_atm.py
```diff
@@ -0,0 +1,162 @@
...
+    """Move the updated dictionary to the atm information file.
+
+    Args:
+        file_path (str): path to atm information file.
+        atm_info (dict): the dictionary, containing all the atm
+                               information.
+    """
+    try:
+        with open(file_path, 'w') as file_pointer:
+            for key in atm_info:
```
rfire01: why not atm_info.items?

rfire01 (changes requested):

exercises/ex_atm.py
shiravak: in one of the reviews igal wrote that when i use -1 i need to write it as a constant, in case it will change to stop or something like it. if the user just want to get out he enter -1 the atm dict will be none and the code will stop.

exercises/ex_atm.py
shiravak: it will turn to minus-  so if my balance is 0, and i withdraw 20$ my balnace will be -20$.
in the exercise they said the balnce can be negative.

exercises/ex_atm.py
rfire01: its ok to stop the program, but no need to change the dict to none

exercises/ex_atm.py
```diff
@@ -0,0 +1,164 @@
...
+
+    The lines of the file will contain user's id, balance and password,
+    with comma's separating them.
+
+    Args:
+        file_path (str): path to atm information file.
+
+    Returns:
+        dict. with user's id as keys,
+            and user's password and balance in another dict as values.
```
jhb1996, line 20: When one dictionary (or list or anything else) is in another the term for that is "nested." Your comment is therefore better expressed as "in a nested dict as values."

Its clearer what you mean this way

exercises/ex_atm.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    try:
+        atm_options = {}
+        with open(file_path, 'r') as file_pointer:
+            for line in file_pointer:
+                line_list = line.rstrip().split(',')
```
jhb1996, line 26: Its our convention to avoid using types ("list", "dict", "string") in  variable names. Better to add more information in the name or just leave it as something like "lines"

PR #129, opened 2020-04-01
Feature/solved atm

lielya:


solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
+"""Simulating an ATM. """
+
+import sys
+
+ATM_FILE = 1
+DICTIONARY = {}
```
natsala13: Using global variable is not a good habit, it makes your code way more difficult to keep trace of, and it could lead to a lot of mistakes since now, every method you write using one of these variables depends not only on its variables but also on the code 'state'.

Remove it, keep in the main function all the variables you need and send them to every method as arguments.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def check_the_balance(costumer_id_st):
+    """Check the state of costumer's money in the bank.
+
+    Args:
+    costumer_id_st(string): costumer id.
+    """
+    balance = ''
+    for i in DICTIONARY.keys():
```
natsala13: Try using more meaningful names to your variables. the name i became common for iterating over numbers as and index. Since here you are iterating over a dictionary.....

natsala13: For iterating over a dictionary you don't one to specify dictionary.keys(). dictionary is enough.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
+"""Simulating an ATM. """
+
+import sys
+
+ATM_FILE = 1
+DICTIONARY = {}
+
+
+def check_the_balance(costumer_id_st):
```
natsala13: Its better not writting the type of a variable in its name. Python is a dynamic language (google it :) ).

solved_ATM_ex/solved_ATM_ex.py
natsala13: When you name a variable, try naming it according to its purpose. DICTIONNARY is a type, it doesn't give me any information about why is that variable here.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def cash_withdrawal(costumer_id_st, amount_to_withdrawal):
+    """Withdrawal from the ATM amount of money.
+
+    Args:
+    costumer_id_st(string): costumer id.
+    amount_to_withdrawal(integer): amount the costumer wants to withdrawal.
+    """
+    balance_now = int(check_the_balance(costumer_id_st))
```
natsala13: It looks like every time you are using the method check_the_balance, you convert its output to int. Why not changing the method itself to return an int?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
...
+            for line in atm_file:
+                list_details = line.split()[1:]
+                DICTIONARY[line.split()[0]] = list_details
+
+        print("welcome to bank!\n")
+        costumer_id = int(input("enter your id:\n"))
+
+        while costumer_id != -1:
+            costumer_id_st = str(costumer_id)
+            if costumer_id_st.isdigit() and len(costumer_id_st) == 9 and \
```
natsala13: why checking is_digit? Is it possible at this part of the program that costumer_id_st is not a number's string?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
...
+                else:
+                    print("choose 1/2/3/4 ")
+                costumer_id = int(input("enter your id:\n"))
+            else:
+                costumer_id = int(input("the id isn't exists/isn't right, "
+                                        "enter a new one:\n"))
+
+        with open(sys.argv[ATM_FILE], 'w') as atm_file:
+            for i in DICTIONARY:
+                atm_file.write(f"{i} {' '.join(DICTIONARY[i])}\n")
```
natsala13: There should be a better way of writing a dictionary to a file in python.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
...
+
+    except FileNotFoundError:
+        print("Your file path is incorrect. Please check it.")
+    except ValueError:
+        print("Value Error! you need to enter an int value")
+    except Exception as e:
+        print(e)
```
natsala13: Try except are excellent, but its not a good habit wrapping all of your code in try, if there a section that might not work, like opening a file, wrap only those lines with a try except.

natsala13: If you got an exception you wasn't expecting, let it flow.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
...
+                return "password changed"
+    return "The password needs to be four digits."
+
+
+def main():
+    try:
+        with open(sys.argv[ATM_FILE], 'r') as atm_file:
+            for line in atm_file:
+                list_details = line.split()[1:]
+                DICTIONARY[line.split()[0]] = list_details
```
natsala13: what is list_details? It seems to be a tuple containing user password and amount of cash in bank. Try using a more meaningful way of storing that data. (dictionary of dictionaries.....)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
...
+    Args:
+    costumer_id_st(string): costumer id.
+    amount_to_withdrawal(integer): amount the costumer wants to withdrawal.
+    """
+    balance_now = int(check_the_balance(costumer_id_st))
+    if balance_now >= amount_to_withdrawal:
+        new_balance = balance_now - amount_to_withdrawal
+        for i in DICTIONARY:
+            if costumer_id_st in i:
+                DICTIONARY[i][1] = str(new_balance)
```
natsala13: It seems you are iterating over the dictionary only to find the user. if you have its id, just access it through the dictionary.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
...
+    Args:
+    costumer_id_st(string): costumer id.
+    amount_to_deposit(integer): amount the costumer wants to deposit.
+    """
+    if amount_to_deposit > 0:
+        balance_now = int(check_the_balance(costumer_id_st))
+        new_balance = balance_now + amount_to_deposit
+        for i in DICTIONARY:
+            if costumer_id_st in i:
+                DICTIONARY[i][1] = str(new_balance)
```
natsala13: Same. the for loop is useless.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
...
+        return f"you have now {new_balance}"
+    return "the amount is not good"
+
+
+def change_password(costumer_id_st, new_password):
```
natsala13: As a general comment, try not returning string messages from methods. If a method verify the cash amount in bank, return it as int and not as a message, the in the main program print whatever you want. If you want to execute a process and return whether it worked or not, return a bool or raise an exception if it didn't worked. 
Anyway don't return string messages from methods.

natsala13 (changes requested):

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,130 @@
+"""Simulating an ATM. """
+
+import sys
+
+ATM_FILE = 1
```
natsala13: Use more meaningful names, 1 is not representing a file.

natsala13 (changes requested):

solved_ATM_ex/solved_ATM_ex.py
lielya: The number 1 refers to the place of the file as a parameter. I didn't understand how can I give a meaningful name instead.

natsala13: "the place of the file as a parameter"
Use these words :)

lielya: You mean like this: 
THE_PLACE_OF_THE_FILE_AS_A_PARAMETER = 1
or maybe: 
ATM_FILE =1  # the place of the file as a parameter

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+
+ATM_FILE = 1  # The number 1 refers to the place of the file as a parameter
+
+
+def check_the_balance(costumer_id, costumers_dictionary):
+    """Check the state of costumer's money in the bank.
+
+    Args:
+    costumer_id (string): costumer id.
+    costumers_dictionary (dictionary): dictionary of all costumers in
```
rfire01: indent (everywhere)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+                                       the bank.
+    """
```
rfire01, line 45: returns? (everywhere)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
+"""Simulating an ATM. """
+
+import sys
+
+ATM_FILE = 1  # The number 1 refers to the place of the file as a parameter
+
+
+def check_the_balance(costumer_id, costumers_dictionary):
```
rfire01: This seem like redundant function, all it do is return a value from dict

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+    """Check the state of costumer's money in the bank.
+
+    Args:
+    costumer_id (string): costumer id.
+    costumers_dictionary (dictionary): dictionary of all costumers in
+                                       the bank.
+    """
+    balance = int(costumers_dictionary[costumer_id][1])
```
rfire01: balance can't be float?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def cash_withdrawal(costumer_id, amount_to_withdrawal, costumers_dictionary):
+    """Withdrawal from the ATM amount of money.
+
+    Args:
+    costumer_id (string): costumer id.
+    amount_to_withdrawal (integer): amount the costumer wants to withdrawal.
+    costumers_dictionary (dictionary): dictionary of all costumers in
+                                       the bank.
```
rfire01: its obvious its dictionary from the type, the text should explain what this dictionary represents (for example bank customer data)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+    """
+    balance_now = check_the_balance(costumer_id, costumers_dictionary)
+    if amount_to_withdrawal > 0:
+        if balance_now >= amount_to_withdrawal:
```
rfire01: write 0 < withdrawal <= balance

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+            new_balance = balance_now - amount_to_withdrawal
+            costumers_dictionary[costumer_id][1] = str(new_balance)
+            return True
```
rfire01: Make sure that everywhere in the code, and any code you write next to have blank line at the end of logic block (if, else, try, etc.)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+        return True
+    return False
+
+
+def main():
+    costumers_dictionary = {}
+    try:
+        with open(sys.argv[ATM_FILE], 'r') as atm_file:
+            for line in atm_file:
+                costumers_dictionary[line.split()[0]] = line.split()[1:]
```
rfire01: why split twice?
you can write something like a, *b = line.split()

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+    """Changing the costumer's ATM password.
+
+    Args:
+    costumer_id (string): costumer id.
+    new_password (string): the new password the costumer wants to change to.
+    costumers_dictionary (dictionary): dictionary of all costumers in
+                                       the bank.
+    """
+    if len(new_password) == 4 and new_password.isdigit():
+        costumers_dictionary[costumer_id][0] = new_password
```
rfire01: This isn't a good representation.
To get to the information you use index to get to the data, but for someone who reads the code, its not clear what he looks at.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+        return True
+    return False
+
+
+def main():
```
rfire01: Whenever you write a code notice that functions should never be more then 15 lines (20 max if there is no choice).
(There are some exception, but generally avoid it whenever you can).
The reason is that if your function more then 20 lines, then probably your function do more then one functionality, and it mean you can split it to smaller functions to increase the readability

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+                        print("not a good password, needs to be 4 digits:")
+
+                else:
+                    print("choose 1/2/3/4 option: ")
+            else:
+                costumer_id = input("the id isn't exists/isn't right, "
+                                    "enter a new one:\n")
+
+    except ValueError:
+        print("Value Error! your input needs to be digit value")
```
rfire01: if someone do a mistake you just close the program?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+
+    except FileNotFoundError:
+        print("Your file path is incorrect. Please check it.")
+
+    try:
+        print("welcome to bank!\n")
+        costumer_id = input("enter your id:\n")
+        while costumer_id != str(-1):
+            if len(costumer_id) == 9 and costumer_id.isdigit() and \
+                    costumers_dictionary.get(costumer_id):
```
rfire01: why do you need to check if the input is a number?
why do you use get function here?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+                print("for cash deposit press 3")
+                print("for change password press 4")
+                costumer_choice = int(input("pressing: \n"))
+
+                if costumer_choice == 1:
+                    print("your balance: ", end='')
+                    print(check_the_balance
+                          (costumer_id, costumers_dictionary))
+                    costumer_id = input("enter your id/ to stop and save"
+                                        " changes press (-1):\n")
```
rfire01: this code duplication, it appear in all of your ifs

rfire01: instead of writing the ID again to continue, why not:
"enter action 1-4, press -1 to exit..."

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+
+                elif costumer_choice == 2:
+                    amount_to_withdrawal = int(input("please enter the "
```
rfire01: Can't I withdraw 1.5 NIS?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+                                                     "amount to withdrawal:\n"
+                                                     ""))
```
rfire01: You can write it look better

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+                    if cash_withdrawal(costumer_id, amount_to_withdrawal,
+                                       costumers_dictionary):
+                        print("The action was succeeded")
+                        costumer_id = input("enter your id/ to stop and save"
+                                            " changes press (-1):\n")
+                    else:
+                        print("failed - check your amount\n")
```
rfire01: if action failed, you don't ask the customer to choose action?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+
+                elif costumer_choice == 3:
+                    amount_to_deposit = int(input("please enter "
```
rfire01: same

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+                                                  "the amount you want"
+                                                  " to deposit: \n "))
+                    if cash_deposit(costumer_id, amount_to_deposit,
+                                    costumers_dictionary):
+                        print("The action was succeeded")
+                        costumer_id = input("enter your id/ to stop and save"
+                                            " changes press (-1):\n")
+                    else:
+                        print("failed - check your amount\n")
```
rfire01: same

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+            for line in atm_file:
+                costumers_dictionary[line.split()[0]] = line.split()[1:]
+
+    except FileNotFoundError:
+        print("Your file path is incorrect. Please check it.")
+
+    try:
+        print("welcome to bank!\n")
+        costumer_id = input("enter your id:\n")
+        while costumer_id != str(-1):
```
rfire01: Anyone can do any action he wants with  only knowing the id? its very easy to steal in this atm

solved_ATM_ex/solved_ATM_ex.py
rfire01: I already made a comment about it in STAM exercise, the purpose of the CR is to learn (and avoid repeating mistakes) not just to finish with it.

rfire01 (changes requested):

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,151 @@
...
+    if amount_to_withdrawal > 0:
+        if balance_now >= amount_to_withdrawal:
+            new_balance = balance_now - amount_to_withdrawal
+            costumers_dictionary[costumer_id][1] = str(new_balance)
+            return True
+    return False
+
+
+def cash_deposit(costumer_id, amount_to_deposit, costumers_dictionary):
+    """deposit to the bank some of money.
```
jhb1996: Capitalize the beginning of every sentence

solved_ATM_ex/solved_ATM_ex.py
jhb1996: its simpler to write '-1' rather than str(-1)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            print("Your ID or your password is not correct. Please try to "
+                  "enter again:\n")
+            costumer_indent = input("Enter your id: ")
+            costumer_password = input("Enter your password: ")
+
+        with open(sys.argv[ATM_FILE], 'w') as atm_file:
+            for line in costumers_dictionary:
+                atm_file.write(
+                    f"{line} {' '.join(costumers_dictionary[line])}\n")
```
jhb1996: The whole reason to use fstrings is that they allow you to write clean readable strings. You don't need to use join in there. Just having a space between your brackets means that there is a space between them.

You should make sure to check your outputs since from the code you have here it looks like you are adding two spaces instead of one which is the kind of thing you should try to catch by looking at the outputs.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
+"""Simulating an ATM. """
+
+import sys
+
+ATM_FILE = 1  # The number 1 refers to the place of the file as a parameter.
+COSTUMER_BALANCE_PLACE_IN_DICTIONARY = 1
```
jhb1996: We avoid putting the name of types in variable/constant names. Instead its better to either be more specific or just leave the string without  the type since the type doesn't add much (change this for the dictionary name as well).

see here for a brief guide to our code style.
https://github.com/osherdp/python-training/blob/master/Style_Guide.md

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+COSTUMER_PASSWORD_PLACE_IN_DICTIONARY = 0
```
rfire01: reorder

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+                    costumers_dictionary):
+    """Withdrawal from the ATM amount of money.
+
+    Args:
+        costumer_indent (string): Costumer id.
+        amount_to_withdrawal (float): Amount the costumer wants to withdrawal.
+        costumers_dictionary (dictionary): Contains bank costumer data.
+
+    Returns:
+        True for success withdrawal action, False otherwise.
```
rfire01: type?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+
```
rfire01, line 41: remove

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+    Returns:
+        A full dictionary of bank costumer data, or an exception if the
+        file path is not correct.
+
+    """
+    try:
+        with open(sys.argv[ATM_FILE], 'r') as atm_file:
+            for line in atm_file:
+                a, *b = line.split()
+                costumers_dictionary[a] = b
```
rfire01: This is a bad practice to save all the data in a list

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+def fill_dictionary(costumers_dictionary):
+    """Filling dictionary with bank customer data.
+
+    Args:
+         costumers_dictionary (dictionary): An empty dictionary.
+
+    Returns:
+        A full dictionary of bank costumer data, or an exception if the
+        file path is not correct.
+
```
rfire01, line 123: remove in all docs this redundant empty line and add type to the return value

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+                costumers_dictionary[a] = b
+            return costumers_dictionary
+
+    except FileNotFoundError:
+        return "Your file path is incorrect. Please check it."
+
+
+def main():
+    costumers_dictionary = {}
+    fill_dictionary(costumers_dictionary)
```
rfire01: Its bad practice to create a dict and then fill it inside of a function, instead you should make a function that return a dict

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+    print("Welcome to bank!\n")
+    costumer_indent = input("Enter your id: ")
+    costumer_password = input("Enter your password: ")
+    not_finish_actions = True
+
+    while not_finish_actions:
+        if costumer_indent in costumers_dictionary and \
```
rfire01: customer_indent? what?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+                    COSTUMER_PASSWORD_PLACE_IN_DICTIONARY] \
+                == costumer_password:
+
+            try:
+                print("For checking your balance press 1")
+                print("For cash withdrawal press 2")
+                print("For cash deposit press 3")
+                print("For change password press 4")
+                print("To exit press -1")
+                costumer_choice = int(input("Pressing: \n"))
```
rfire01: you can save the casting and just compare strings

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+
+                if costumer_choice == 1:
+                    print(f"Your balance: "
+                          f"{float(costumers_dictionary[costumer_indent][COSTUMER_BALANCE_PLACE_IN_DICTIONARY])}\n")
```
rfire01: pep8 says that line shouldn't pass the 100 char length

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+
+def main():
+    costumers_dictionary = {}
+    fill_dictionary(costumers_dictionary)
+    print("Welcome to bank!\n")
+    costumer_indent = input("Enter your id: ")
+    costumer_password = input("Enter your password: ")
+    not_finish_actions = True
+
+    while not_finish_actions:
```
rfire01: At the moment once the user finish, you program terminates instead of letting the next customer work (the task says terminate the program when -1 entered)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,180 @@
...
+            for line in atm_file:
+                a, *b = line.split()
+                costumers_dictionary[a] = b
+            return costumers_dictionary
+
+    except FileNotFoundError:
+        return "Your file path is incorrect. Please check it."
+
+
+def main():
```
rfire01, line 162: This function too long.
I will make it easier for this time, in your next submission you are not allowed to pass the 25 lines in the main function

rfire01 (changes requested):

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,156 @@
+"""Simulating an ATM. """
+
+import sys
+import json
+
+ATM_FILE = 1  # The number 1 refers to the place of the file as a parameter.
```
rfire01: Usually your code need to be simple enough that you don't need doc to explain some line.
Sometime its ok to explain if there is complex logic section, however, variable naming should always be clear without doc

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,156 @@
...
+def create_dictionary(path_of_file):
+    """Create dictionary of bank customer data.
+
+    Args:
+        path_of_file (string): Path of the file that contains all bank
+        costumer data.
+
+    Returns:
+        dictionary: A nested dictionary of bank costumer data, or an exception
+        if the file path is not correct.
```
rfire01: if you got a variable \ return value in the doc that require more then a single line, the next line should be with indent (for example line 72 "customer data" should be with indent)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,156 @@
...
+                        costumer_password = new_password
+                        print("Password changed")
+
+                    else:
+                        print("Not a good password,it needs to be 4 digits.")
+
+                elif costumer_choice == 5:
+                    costumer_id = input("Enter your id: ")
+                    if costumer_id != '-1':
+                        costumer_password = input("Enter your password: ")
```
rfire01: Exit code is -1, not 6

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,156 @@
...
+            data = json.load(file_j)
+            costumers_data = data
+            return costumers_data
+
+    except FileNotFoundError:
+        return "Your file path is incorrect. Please check it."
+
+
+def main():
+    costumers_data = create_dictionary(path_of_file=sys.argv[ATM_FILE])
```
rfire01: what happen if path is not legal?
your program is going to crash

rfire01 (changes requested):

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -69,30 +70,30 @@ def create_dictionary(path_of_file):
...
+    if os.path.isfile(path_of_file):
         with open(path_of_file, 'r') as file_j:
             data = json.load(file_j)
             costumers_data = data
             return costumers_data
 
-    except FileNotFoundError:
-        return "Your file path is incorrect. Please check it."
+    print("Your file path is incorrect. Please check it.")
+    quit()
```
rfire01: You almost never want to run "quit" in a function.
Think about that you are working with calculator, and if you tried to divide by 0 the calculator will close

rfire01 (changes requested):

rfire01:
reminding you to click on resolve on any comment you fixed

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
+"""Simulating an ATM. """
+
```
rfire01: remove

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+ATM_FILE_PLACE = 1
+COSTUMER_CHOICE_OPTIONS = "For checking balance press  1 \nFor cash " \
+                          "withdrawal press 2 \nFor cash deposit press 3 " \
+                          "\nFor change password " \
+                          "press 4 \nTo finis actions press 5 \n"
+
+
+def cash_withdrawal(costumer_id, costumers_data):
+    """Withdrawal from the ATM amount of money.
```
rfire01: first word should be command

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+        balance_now = float(costumers_data[costumer_id]['balance'])
+        new_balance = balance_now + amount_to_deposit
+        costumers_data[costumer_id]['balance'] = new_balance
+        return True
+
+    return False
+
+
+def change_password(costumer_id, costumers_data, new_password):
+    """Changing costumer's ATM password.
```
rfire01: command word

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+    if os.path.isfile(path_of_file):
+        with open(path_of_file, 'r') as file_j:
+            data = json.load(file_j)
+            costumers_data = data
+            return costumers_data
+
+    print("File not found")
+    costumers_data = None
+    return costumers_data
```
rfire01: just return None

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+    Args:
+        costumer_choice (integer): Costumer choice from 1/2/3/4/5 options,
+            references to the action he wants to do in ATM.
+        costumers_data (dictionary): Bank costumer data.
+        costumer_id (string): Costumer id.
+        costumer_password (string): Current costumer password.
+
+    Returns:
+        string: Updated password if the choice is 4, or updated id and
+            password if the choice is 5.
```
rfire01: This is not correct.
Your function can return either None, string or tuple.
Your function should return only one thing

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+        new_password = input("New password: ")
+        costumer_password = new_password if change_password(
+            costumer_id, costumers_data, new_password) else costumer_password
+        return costumer_password
+
+    elif costumer_choice == 5:
+        costumer_id = input("id: ")
+        if costumer_id != '-1':
+            costumer_password = input("Password: ")
+        return costumer_id, costumer_password
```
rfire01: Try to think on a better way to implement it.
Do you really care if the actions succeeded or not?

rfire01: This function suppose to handle the users actions.
The functionality for getting a new user, isn't part of a user actions

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+def manage_costumer_actions(costumer_id, costumers_data, costumer_password):
+    """Manage costumer actions.
+
+     Args:
+        costumers_data (dictionary): Bank costumer data.
+        costumer_id (string): Costumer id.
+        costumer_password (string): Current costumer password.
+
+    Returns:
+        None.
```
rfire01: If you don't return anything this, you don't need to write this section

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+                                                              costumer_password)
+
+                elif costumer_choice == 5:
+                    costumer_id, costumer_password = check_costumer_choice(
+                        costumer_choice, costumers_data, costumer_id,
+                        costumer_password)
+
+                else:
+                    check_costumer_choice(costumer_choice, costumers_data,
+                                          costumer_id, costumer_password)
```
rfire01: This is bad design.
Since your function return different things for different inputs you have to check how to save it.
You should not check the customer_choice twice, both here and in the function.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+
+            except ValueError:
+                print("Value Error! your input needs to be digit value\n")
```
rfire01: Why all of this in try except?
From the look of it you only need it on the conversion to int.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+        else:
+            print("ID or password is not correct.Try to enter again:\n")
+            costumer_id, costumer_password = input("id: "), input(
+                "Password: ")
+
+
+def main():
+    costumers_data = create_dictionary(path_of_file=sys.argv[ATM_FILE_PLACE])
+    if costumers_data is not None:
+        costumer_id, costumer_password = input("id: "), input("Password: ")
```
rfire01: Think again about the design of your code.
You read the id and password in 3 different places, is it really necessary?

rfire01 (changes requested):

solved_ATM_ex/solved_ATM_ex.py
jhb1996: To clarify, he means that the first word of any function doc should be "present progressive." (this generally means a verb ending in an 's')

For some made up examples:
"""Checks that two numbers are equal"""
"""Sends a message to the server"""
"""Reformats the list to the standard format"""

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,190 @@
...
+        costumer_id (string): Costumer id.
+        costumers_data (dictionary): Bank costumer data.
+        new_password (string): Password the costumer wants to change to.
+
+    Returns:
+        boolean: True for success changing the password, False otherwise.
+    """
+    if len(new_password) == 4 and new_password.isdigit():
+        costumers_data[costumer_id]['password'] = new_password
+        password_now = new_password
```
jhb1996: There is a bug here. Look carefully. You are not returning a boolean.

solved_ATM_ex/solved_ATM_ex.py
rfire01: its correct except for the end with 's'
in the examples, it should be check, send or reformat

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    switcher = {
+        '1': lambda: print(costumers_data[costumer_id]['balance']),
+        '2': lambda: cash_withdraw(costumers_data),
+        '3': lambda: cash_deposit(costumers_data),
+        '4': lambda: change_password(costumers_data),
+        '5': lambda: main()
+    }
+
+    return switcher.get(costumer_choice, lambda: print("Invalid option"))()
```
IgalKolihman: Beautiful solution, I think it's very original 👍🏼

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+
+    global costumer_id, costumer_password
+    costumer_id = input("id: ")
+    if costumer_id != '-1':
+        costumer_password = input("Password: ")
+
+
+def check_costumer_choice(costumer_choice, costumers_data):
+    """Check costumer's choice and act accordingly.
+
```
IgalKolihman, line 123: add a description of the available choices

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    return switcher.get(costumer_choice, lambda: print("Invalid option"))()
+
+
+def manage_costumer_actions(costumers_data):
+    """Manage costumer actions.
+
+     Args:
+        costumers_data (dictionary): Bank costumer data.
```
IgalKolihman: this is not how you write the 'dictionary' type

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+    costumer_id = input("id: ")
+    if costumer_id != '-1':
+        costumer_password = input("Password: ")
+
+
+def check_costumer_choice(costumer_choice, costumers_data):
+    """Check costumer's choice and act accordingly.
+
+    Args:
+        costumer_choice (string): Costumer choice from 1/2/3/4/5 options,
```
IgalKolihman: this is not how you represent a type of 'string'

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+            costumers_data = data
+            return costumers_data
+
+    print("File not found")
+    return None
+
+
+def get_new_costumer():
+    """Get a new costumer. """
+
```
IgalKolihman: remove line

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+def create_dictionary(path_of_file):
+    """Create dictionary of bank customer data.
+
+    Args:
+        path_of_file (string): Path of file that contains all bank costumer
+            data.
+
+    Returns:
+        dictionary: A nested dictionary of bank costumer data, or None.
```
IgalKolihman:
```suggestion
        dict: A nested dictionary of bank customer data, or None.
```

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+    if len(new_password) == 4 and new_password.isdigit():
+        costumers_data[costumer_id]['password'] = new_password
+        costumer_password = new_password
+        return True
+
+    print("Should be 4 digits.")
+    return False
+
+
+def create_dictionary(path_of_file):
```
IgalKolihman: rename

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+    Returns:
+        boolean: True for success deposit action, False otherwise.
+    """
+    amount_to_deposit = input("Amount to deposit: ")
+    check_amount = check_convert_to_float(amount_to_deposit)
+    if check_amount is not None:
+        if check_amount > 0:
+            balance_now = float(costumers_data[costumer_id]['balance'])
+            new_balance = balance_now + check_amount
+            costumers_data[costumer_id]['balance'] = new_balance
```
IgalKolihman: set `'balance'` and `'password'` into CONSTANT variables. This will make your life easier in future changes because you are only will have to change in one place.

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+
+    Args:
+        costumers_data (dictionary): Bank costumer data.
+
+    Returns:
+        boolean: True for success deposit action, False otherwise.
+    """
+    amount_to_deposit = input("Amount to deposit: ")
+    check_amount = check_convert_to_float(amount_to_deposit)
+    if check_amount is not None:
```
IgalKolihman:
```suggestion
    check_amount = check_convert_to_float(amount_to_deposit)
    if check_amount:
```

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+
+    Args:
+        costumers_data (dictionary): Bank costumer data.
+
+    Returns:
+        boolean: True for success withdraw action, False otherwise.
+    """
+    amount_to_withdrawal = input("Amount to withdrawal: ")
+    check_amount = check_convert_to_float(amount_to_withdrawal)
+    if check_amount is not None:
```
IgalKolihman:
```suggestion
    if check_amount:
```

IgalKolihman (changes requested):
Greate code! after fixing these last things I believe you can continue to the next exercise. Good luck 👍🏼

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
+import os
+import sys
+import json
+
+ATM_FILE_PLACE = 1
+COSTUMER_CHOICE_OPTIONS = "1 - View balance \n2 - Withdraw \n3 - Deposit " \
+                          "\n4 - Change password \n5 - Exit\n"
```
rfire01: it will look nicer if you actually splitted it to lines

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+
+costumer_id = ""
+costumer_password = ""
```
rfire01: try to avoid as much as possible from global parameters

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+    Args:
+        amount (string): Costumer's input.
+
+    Returns:
+        float: Change the type of the variable 'amount' and return it,
+            else if couldn't- return None.
+    """
+    try:
+        convert_type = float(amount)
+        return convert_type
```
rfire01: what about blank line after logic block? :(

rfire01: enough to do
return float(amount)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+def change_password(costumers_data):
+    """Change costumer's password.
+
+    Args:
+        costumers_data (dictionary): Bank costumer data.
+
+    Returns:
+        boolean: True for success changing the password, False otherwise.
+    """
+    global costumer_password
```
rfire01: avoid globals

solved_ATM_ex/solved_ATM_ex.py
rfire01: What he means,  that its not clear from the name what the function do.
function create_dictionary can return for example just {}, it says nothing about the content

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+    Returns:
+        string: An error message if costumer's choice is not in range.
+    """
+    switcher = {
+        '1': lambda: print(costumers_data[costumer_id]['balance']),
+        '2': lambda: cash_withdraw(costumers_data),
+        '3': lambda: cash_deposit(costumers_data),
+        '4': lambda: change_password(costumers_data),
+        '5': lambda: main()
+    }
```
rfire01: Much better!
There is a simple way (without changing a lot) to make this a const, try to think how to do it

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+
+    Returns:
+        string: An error message if costumer's choice is not in range.
+    """
+    switcher = {
+        '1': lambda: print(costumers_data[costumer_id]['balance']),
+        '2': lambda: cash_withdraw(costumers_data),
+        '3': lambda: cash_deposit(costumers_data),
+        '4': lambda: change_password(costumers_data),
+        '5': lambda: main()
```
rfire01: This doesn't make sense, if someone finish to run all command he wants, you run main again?
after some customers you will get nested functions to big depth

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+    costumers_data = create_dictionary(path_of_file=sys.argv[ATM_FILE_PLACE])
+    if costumers_data is not None:
+        manage_costumer_actions(costumers_data)
+
+        with open(sys.argv[ATM_FILE_PLACE], 'w') as atm_file:
+            atm_file.write(json.dumps(costumers_data))
+
+
+if __name__ == '__main__':
+    main()
```
rfire01, line 170: The code looks much better, good job

rfire01 (changes requested):

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,174 @@
...
+    check_amount = check_convert_to_float(amount_to_withdrawal)
+    if check_amount is not None:
+        balance_now = float(costumers_data[costumer_id]['balance'])
+        if 0 < check_amount <= balance_now:
+            new_balance = balance_now - check_amount
+            costumers_data[costumer_id]['balance'] = new_balance
+            return True
+
+    print("Failed - check amount\n")
+    return False
```
rfire01: I all functions you return True\False, but when you use them you don't do anything with what they return, so do you really need it?

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -3,103 +3,103 @@
...
+                          "3 - Deposit \n" \
+                          "4 - Change password \n" \
+                          "5 - Exit\n"
+SWITCHER = {
+    '1': lambda data, identifier: print(data[identifier][
+                                            BALANCE_OF_COSTUMER]),
+    '2': lambda data, identifier: cash_withdraw(data, identifier),
+    '3': lambda data, identifier: cash_deposit(data, identifier),
+    '4': lambda data, identifier: change_password(data, identifier)
+}
```
jhb1996, line 19: In general we don't like lambdas and prefer you just use helper functions. Its easier to read the code that way.

for 2-4 the lambdas are completely unnecessary as you can simply use the function without any arguments as the value.

Here is an example of what I mean:
def f_1(a,b):
    print(a+b)

def f_2(a,b):
    print(a-b)

MATCHER = {1:f_1, 2:f_2}
result_1 = MATCHER[1](1,2) # equals 3

result_2 = MATCHER[2](1,2) #equals -1

print(result_1, result_2)

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -111,63 +111,59 @@ def create_dictionary(path_of_file):
...
 def main():
-    get_new_costumer()
-    costumers_data = create_dictionary(path_of_file=sys.argv[ATM_FILE_PLACE])
-    if costumers_data is not None:
-        manage_costumer_actions(costumers_data)
-
-        with open(sys.argv[ATM_FILE_PLACE], 'w') as atm_file:
-            atm_file.write(json.dumps(costumers_data))
+    costumers_data = read_costumers_information_from_file(
+        path_of_file=sys.argv[ATM_FILE_PLACE])
```
jhb1996, line 164: leave spaces before and after the equal sign.

jhb1996 (changes requested):

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,170 @@
...
+            data.
+
+    Returns:
+        dict: A nested dictionary of bank costumer data, or None.
+    """
+    if os.path.isfile(path_of_file):
+        with open(path_of_file, 'r') as file_j:
+            data = json.load(file_j)
+            costumers_data = data
+            return costumers_data
```
IgalKolihman, line 108: Why not just:
`return data`

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,170 @@
...
+def manage_costumer_actions(costumers_data, costumer_id, costumer_password):
+    """Manage costumer actions.
+
+     Args:
+        costumers_data (dict): Bank costumer data.
+        costumer_id (str): Costumer's ID.
+        costumer_password (str): Costumer's password.
+    """
+    if costumer_id in costumers_data and costumers_data[
+            costumer_id][PASSWORD_OF_COSTUMER] == costumer_password:
```
IgalKolihman, line 133: The `if` is too long

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,170 @@
...
+                "Invalid option"))(costumers_data, costumer_id)
+            costumer_choice = input(COSTUMER_CHOICE_OPTIONS)
+
+        print("Thank you.")
+
+    else:
+        print("ID or password is not correct. Try again:\n")
+
+
+def manage_new_costumer(costumers_data):
```
IgalKolihman, line 146: I think that this function does something more like a login. Rename.

IgalKolihman (changes requested):
Inform me when you done 👍🏻

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,170 @@
...
+                          "2 - Withdraw \n" \
+                          "3 - Deposit \n" \
+                          "4 - Change password \n" \
+                          "5 - Exit\n"
+SWITCHER = {
+    '1': lambda data, identifier: print(data[identifier][
+                                            BALANCE_OF_COSTUMER]),
+    '2': lambda data, identifier: cash_withdraw(data, identifier),
+    '3': lambda data, identifier: cash_deposit(data, identifier),
+    '4': lambda data, identifier: change_password(data, identifier)
```
rfire01, line 18: What about what we talked about?
both lines do exactly the same:
`res1 = lambda x: func(x)`
`res2 = func`
Both lines equal - res1 = res2

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,170 @@
...
+     Args:
+        costumers_data (dict): Bank costumer data.
+        costumer_id (str): Costumer's ID.
+        costumer_password (str): Costumer's password.
+    """
+    if costumer_id in costumers_data and costumers_data[
+            costumer_id][PASSWORD_OF_COSTUMER] == costumer_password:
+        costumer_choice = input(COSTUMER_CHOICE_OPTIONS)
+        while costumer_choice != '5':
+            SWITCHER.get(costumer_choice, lambda data, identifier: print(
```
rfire01, line 136: if you don;t use the lambda args, you don't need to give them names, you can do one of the two:
lambda _,_: print()
lambda *_: print()

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,170 @@
...
+    """Change costumer's password.
+
+    Args:
+        costumers_data (dict): Bank costumer data.
+        costumer_id (str): Costumer's ID.
+    """
+    new_password = input("New password: ")
+    if len(new_password) == 4 and new_password.isdigit():
+        costumers_data[costumer_id][PASSWORD_OF_COSTUMER] = new_password
+        update_changes_in_dictionary(costumers_data)
```
rfire01, line 88: This is a bad way to do it.
Think about the fact that in real life you got hundreds of thousands of customers, if you need to update the entire DB for each input you will huge delays

solved_ATM_ex/solved_ATM_ex.py
```diff
@@ -0,0 +1,170 @@
...
+def change_password(costumers_data, costumer_id):
+    """Change costumer's password.
+
+    Args:
+        costumers_data (dict): Bank costumer data.
+        costumer_id (str): Costumer's ID.
+    """
+    new_password = input("New password: ")
+    if len(new_password) == 4 and new_password.isdigit():
+        costumers_data[costumer_id][PASSWORD_OF_COSTUMER] = new_password
```
rfire01, line 87: This is a source for bugs.
Its correct that if you change dict in function you will change the source dict, but it make it harder to miss it if there is a bug.
If you got function you want to update a dict inside of it, you should return the updated dict at the end, this way its more clear where you change it, and easier to debug if there is a problem

rfire01 (changes requested):

PR #133, opened 2020-04-02
matrix

yonatan77:


basic/matrix/Matrix_main.py
```diff
@@ -0,0 +1,15 @@
+from Matrix_class import Matrix
+mat = Matrix(((1,2), (3,4)))
+mat2 = Matrix(((1,2), (3,4)))
+mat.tuples()
+print(mat)
+print(mat.unity(4))
+print(mat.ones(6))
+print(mat + mat2.tuples())
+print(mat - mat2.tuples())
```
Liavatrix, line 9: It was supposed to liik like mat + mat2, and not mat+mat2.tuples

PR #139, opened 2020-04-05
Calculator exercise

mayareu:


Atm.py
```diff
@@ -0,0 +1,136 @@
...
+
+from Costumer import Costumer
+
+COSTUMERS_DICTIONARY = {}
+FILE_PATH = r'C:\Users\mayar\Desktop\Python\ATM\costumers.txt'
+CONSUMER = ''
+
+
+def get_current_consumer():
+    consumer_id = raw_input('To turn off insert -1\nPlease enter your id:\n')
```
jhb1996, line 11: Starting with Python 3, raw_input() was renamed to input() (change everywhere)

Costumer.py
```diff
@@ -0,0 +1,17 @@
+class Costumer(object):
```
jhb1996, line 1: doc 

see here for details https://github.com/osherdp/python-training/blob/master/README.md

Costumer.py
```diff
@@ -0,0 +1,17 @@
...
+        self.balance = balance
+
+    def set_password(self, new_password):
+        self.password = new_password
+
+    def get_balance(self):
+        return self.balance
+
+    def set_balance(self, updated_bal):
+        self.balance = updated_bal
```
jhb1996, line 14: Your code treats the balance as a string and converts it to a number every time you want to use it. Better to  store it as a number

Atm.py
```diff
@@ -0,0 +1,136 @@
+from pip._vendor.distlib.compat import raw_input
```
jhb1996, line 1: Pretty sure you accidentally pushed the atm files along with the calculator code

Calculator.py
```diff
@@ -0,0 +1,134 @@
+import re
```
jhb1996, line 1: Doc

Calculator.py
```diff
@@ -0,0 +1,134 @@
+import re
+import math
+from statistics import mean
+
+from pip._vendor.msgpack.fallback import xrange
+
+
+class Calculator:
```
jhb1996, line 8: more doc

Calculator.py
```diff
@@ -0,0 +1,134 @@
...
+                                        4: r'%',
+                                        5: r'\^',
+                                        6: r'[/*]',
+                                        7: r'[-+]'}
+        self.exercise = exercise.strip()
+        self.solved = ''
+        print(self.arrange)
+
+    @property
+    def arrange(self):
```
jhb1996, line 22: every function needs a doc (this isn't just because we are annoying. Im actually not sure what this function does)

Calculator.py
jhb1996, line 1: do you have test cases? If so push those too

Calculator.py
```diff
@@ -0,0 +1,134 @@
...
+        self.solved = re.search(r'(!\d+)|(([~!]\d+)?\d+[-()%^/*+$&@]([~!]\d+)?\d+)', self.exercise)
+        if self.solved is None:
+            return self.exercise
+        exercise = self.exercise
+        for operation in self.__operations_regex_dict.values():
+            if self.__operations_regex_dict[1] == operation:
+                reg = str(operation)
```
jhb1996, line 29: we dont use abreviations for variable names

Calculator.py
```diff
@@ -0,0 +1,134 @@
...
+                                        5: r'\^',
+                                        6: r'[/*]',
+                                        7: r'[-+]'}
+        self.exercise = exercise.strip()
+        self.solved = ''
+        print(self.arrange)
+
+    @property
+    def arrange(self):
+        self.solved = re.search(r'(!\d+)|(([~!]\d+)?\d+[-()%^/*+$&@]([~!]\d+)?\d+)', self.exercise)
```
jhb1996, line 23: turn a big regex like this into a constant. That way it will have a name and i will know what you are searching for

Calculator.py
```diff
@@ -0,0 +1,134 @@
...
+        if self.solved is None:
+            return self.exercise
+        exercise = self.exercise
+        for operation in self.__operations_regex_dict.values():
+            if self.__operations_regex_dict[1] == operation:
+                reg = str(operation)
+                exp = re.search(reg, exercise)
+            elif self.__operations_regex_dict[2] == operation:
```
jhb1996, line 31: blank line between each logic block (at the end of every if, elif, else, for and while)

Calculator.py
```diff
@@ -0,0 +1,134 @@
...
+        self.solved = ''
+        print(self.arrange)
+
+    @property
+    def arrange(self):
+        self.solved = re.search(r'(!\d+)|(([~!]\d+)?\d+[-()%^/*+$&@]([~!]\d+)?\d+)', self.exercise)
+        if self.solved is None:
+            return self.exercise
+        exercise = self.exercise
+        for operation in self.__operations_regex_dict.values():
```
jhb1996, line 27: your logic here is a bit all over the place. Your goal (if I understand correctly) is to handle each operation in order of its priority. However what you seem to be doing  is handling each operation in random order which is a problem.

Feel free to correct me if I am misunderstanding but there is definitely something here that is at a minimum done strangely.

Calculator.py
```diff
@@ -0,0 +1,134 @@
...
+class Calculator:
+    def __init__(self, exercise):
+        self.__operations_regex_dict = {1: r'\(\d+[!%^+/*\-~$&@]\d+\)+',
+                                        2: r'\!',
+                                        3: r'[$&@]',
+                                        4: r'%',
+                                        5: r'\^',
+                                        6: r'[/*]',
+                                        7: r'[-+]'}
+        self.exercise = exercise.strip()
```
jhb1996, line 17: consider a different name instead of exercise because as I read this I'm not sure what exercise means. Did you mean expression?

PR #140, opened 2020-04-05
commit atm

kerenkay:


ATM.py
```diff
@@ -0,0 +1,106 @@
...
+    atm = {}
+    with open(filename, 'r') as file:
+        for line in file:
+            details = line.strip().split(' ')
+            atm[details[0]] = details[1:]
+
+    return atm
+
+
+def atm_machine(id, atm_dic):
```
natsala13: Avoid putting variable type in their names.

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    Args:
+        id (str)- first parameter
+        atm_dic (dic)- second parameter
```
natsala13: dict 😅

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+
+    :return:
+        A dictionary with the updated values
+        after performing the desired action
+
+    """
+    if atm_dic.get(id):
```
natsala13: Not the right way of accessing a dictionary. Google it about how to use a dictionary in python.

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+                       "2- for cash withdrawal\n"
+                       "3- for cash deposit\n"
+                       "4- for change password\n")
+
+        if action == '1':
+            print(f"your balance is: {atm_dic[id][BALANCE]}")
+
+        elif action == '2':
+            money = input("enter the amount you want to withdraw\n")
+            atm_dic[id][BALANCE] = str(int(atm_dic[id][BALANCE]) - int(money))
```
natsala13: What would happen if one withdraw more than what he has?

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+The script simulates ATM that preforms the following:
+balance check, withdraw cash, deposit cash and
+password change.
+
+"""
+import os
+import sys
+
+
+PATH=1
```
natsala13: I would expect to find a path to something in this variable, rename it to something more specific.

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+BALANCE=1
```
natsala13: Great usage of constants!

ATM.py
```diff
@@ -0,0 +1,106 @@
+"""
+
```
natsala13: Remove empty lines.

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+def file_to_dic(filename):
+    """
+
+    :arg
+        filename(.txt file) - parameter
+
+    :return:
+        a dictionary, key=id and value=[password, balance]
+
+    """
```
natsala13:
```suggestion
    """<short description>.
    
    Args:
        filename(.txt file) - parameter.

    Returns:
        a dictionary, key=id and value=[password, balance].
    """
```

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+        while id != '-1':
+            atm_machine(id, atm_dic)
+            id = input('please enter your ID\n')
+
+        with open(sys.argv[PATH],'w') as newfile:
+            for id in atm_dic:
+                newfile.writelines(f"{id} {atm_dic[id][PASSWORD]} {atm_dic[id][PASSWORD]}\n")
+
+    else:
+        print('file not found')
```
natsala13: Raise an exception instead of printing an error.

natsala13: Now that you fixed the comment above and you are throwing an exception, what would happen if you just tried opening the file without checking if it exists first?

natsala13 (changes requested):

ATM.py
jhb1996: Make sure to use spaces between the equal signs. It makes the code more readable (do this everywhere). See section 2.7 https://github.com/osherdp/python-training/blob/master/Style_Guide.md

ATM.py
jhb1996: To clarify, get() works here but if you just want to check if a key is in the dictionary use.

if id in atm_dic: 

since its easier to read.

ATM.py
jhb1996: I can tell that its the first parameter just by looking at the function def. Tell me what the parameters are used for.

ATM.py
```diff
@@ -0,0 +1,106 @@
+"""
+
+The script simulates ATM that preforms the following:
+balance check, withdraw cash, deposit cash and
+password change.
+
+"""
```
IgalKolihman: use the correct format for documentation.
first list must continue from `"""`  and end with a `.`. this line should describe shortly what the script does.

if you have some more description to add, got down 2 lines and write it there

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+PATH=1
+BALANCE=1
+PASSWORD=0
+
+
+def file_to_dic(filename):
+    """
+
+    :arg
+        filename(.txt file) - parameter
```
IgalKolihman:
```suggestion
        filename(str) - file name ending with '.txt'
```

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+
+    :return:
+        a dictionary, key=id and value=[password, balance]
```
IgalKolihman: this is not you describe the return

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+
+    """
+    atm = {}
+    with open(filename, 'r') as file:
+        for line in file:
+            details = line.strip().split(' ')
+            atm[details[0]] = details[1:]
```
IgalKolihman: add in the documentation an explanation of the content format that you work with

ATM.py
IgalKolihman: along with all your documentation sections, change the `-` sign with `:`.
and don't forget to add space between the attribute name and its type.
for example:
```suggestion
        filename (.txt file): parameter
```

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+def main():
+    """
+
+    :arg
+        id(int) - 9 digits ID (example: 322781145)
+
+    :return:
+        will do the desired ATM operation accordig to an
+        ID that has been accepted until receive -1 as an ID.
+        all the chages will be saved to the ATM.
```
IgalKolihman: this should be in the description, not in the return

ATM.py
```diff
@@ -0,0 +1,106 @@
...
+
+        elif action == '4':
+            password = input('enter new password\n')
+            atm_dic[id][PASSWORD] = password
+
+        else:
+            print('action not detected')
+
+    else:
+        print('there is an error i the ID you etered')
```
IgalKolihman: when you print something to the terminal (like logs or system messages) always start the message sentence with a capital letter.

for example:
```suggestion
        print('There is an error i the ID you etered')
```

IgalKolihman: check your grammar

IgalKolihman (changes requested):

rfire01:
make sure flake 8 pass

ATM.py
```diff
@@ -1,51 +1,50 @@
-"""
+"""The script simulates ATM.
 
 The script simulates ATM that preforms the following:
 balance check, withdraw cash, deposit cash and
 password change.
-
 """
```
rfire01: This should be a single line that describe the file

ATM.py
```diff
@@ -71,35 +74,38 @@ def atm_machine(id, atm_dic):
...
+        with open(sys.argv[PATH], 'w') as newfile:
             for id in atm_dic:
-                newfile.writelines(f"{id} {atm_dic[id][PASSWORD]} {atm_dic[id][PASSWORD]}\n")
+                newfile.writelines(f"{id} {atm_dic[id][PASSWORD]} "
+                                   f"{atm_dic[id][BALANCE]}\n")
 
-    else:
-        print('file not found')
+    except Exception as e:
+        print(e)
```
natsala13, line 108: If you catch an exception just to print it, its redundant.
Its ok to let the program fail if there was an error.

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+                print('The account balance is less than the '
+                      'then the amount you wanted to withdraw')
+
+        elif action == '3':
+            money = input("enter the amount you want to deposit\n")
+            atm_dic[id][BALANCE] = str(int(atm_dic[id][BALANCE]) + int(money))
+
+        elif action == '4':
+            password = input('enter new password\n')
+            atm_dic[id][PASSWORD] = password
```
natsala13, line 71: This functions is too long, Split it to smaller methods.
It will make your code clearer and explain what you are doing without even documenting.
Even if you have a small method doing only one line, its not complicating your code, its making it more readble.

natsala13 (changes requested):

ATM.py
```diff
@@ -0,0 +1,112 @@
+"""The script simulates ATM.
+
+The script simulates ATM that preforms the following:
+balance check, withdraw cash, deposit cash and
+password change.
+"""
```
rfire01, line 6: This part should be a single line that describes what the file purpose is

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+        filename (str): file name ending with '.txt'
+
+    Returns:
+        dictionary: key=id and value=[password, balance]
+    """
+    atm = {}
+    with open(filename, 'r') as file:
+        for line in file:
+            details = line.strip().split(' ')
+            atm[details[ID_INDEX]] = details[INDEX_START_OF_PASSWORD_AND_BALANCE:]
```
rfire01, line 31: Saving the data as list isn't a good practice.
What happen if you change the way the data look? you need to change everywhere you use it the indexes of each part.
Instead the only place that should be effect is the read file function

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+
+    return atm
+
+
+def operation_of_atm(id, atm_dic):
+    """Performs the desired ATM operation for the inserted id.
```
rfire01, line 37: Doc need to start with imperative words

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Args:
+        id (str): the id for which you want the ATM to work
```
rfire01, line 40: the atm works without any id. you mean the use id to operate an action for

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+        atm_dic (dict): dictionary containing all ATM information
```
rfire01, line 41: dot at end of sentence (fix it in all docs in the file)

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+    atm = {}
+    with open(filename, 'r') as file:
+        for line in file:
+            details = line.strip().split(' ')
+            atm[details[ID_INDEX]] = details[INDEX_START_OF_PASSWORD_AND_BALANCE:]
+
+    return atm
+
+
+def operation_of_atm(id, atm_dic):
```
rfire01, line 36: name shouldn't contain the variable type

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+        elif action == '2':
+            money = input("enter the amount you want to withdraw\n")
+            if int(atm_dic[id][BALANCE]) > int(money):
+                atm_dic[id][BALANCE] = str(int(atm_dic[id][BALANCE]) - int(money))
+            else:
+                print('The account balance is less than the '
+                      'then the amount you wanted to withdraw')
+
+        elif action == '3':
+            money = input("enter the amount you want to deposit\n")
```
rfire01, line 66: what happen if someone enters a negative number?

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+
+        elif action == '4':
+            password = input('enter new password\n')
+            atm_dic[id][PASSWORD] = password
+
+        else:
+            print('action not detected')
+
+    else:
+        print('There is an error in the ID you entered')
```
rfire01, line 77: you don't check if there is an error in the id, you check if the id exists in the system

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Args:
+        id (str): the id for which you want the ATM to work
+        atm_dic (dict): dictionary containing all ATM information
+
+    Returns:
+        dictionary: dictionary with the updated values
+            after performing the desired action
+    """
+    if id in atm_dic:
```
rfire01, line 47: generally very good job in this function.
Try to go with convention that says if you got a function with more then 15~20 lines of code, then it usually means that you do more then one functionality in this function and you can split it to several smaller functions

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+
+def main():
+    """Do the ATM operation and save the changes
+
+    will do the desired ATM operation according to an
+    ID that has been accepted until receive -1 as an ID.
+    all the changes will be saved to the ATM.
+
+    Arg:
+        id (int): 9 digits ID (example: 322781145)
```
rfire01, line 88: There are no args

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Raises:
+        Exception as e: No such file or directory
+            if the path to the file not exist will be an error.
```
rfire01, line 92: Notice that isn't how you write doc for exception.
1. you write this part of the doc only if __you__ raise an exception
2. if you do raise exception, then it should be explicit what error is it (for example ValueError)

ATM.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+    try:
+        atm_dic = file_to_dic(sys.argv[PATH_TO_FILE])
+        id = input('please enter your ID\n')
+
+        while id != '-1':
+            operation_of_atm(id, atm_dic)
+            id = input('please enter your ID\n')
```
rfire01, line 100: This look like an atm that very easy to steal money from (you never need to enter your password at any stage)

rfire01 (changes requested):

